PR #3, opened 2016-02-05
Feature/implement field method

romaninsh:
Copy and refactor of some methods from dsql()

 - field
 - render_field => _render_field
 - bt => _escape
 - consume => _consume

Includes documentation and additional tests.

screen shot 2016-02-05 at 15 28 05


romaninsh:
Thanks ;)

PR #6, opened 2016-02-08
Public / protected methods, extend unit tests functionality

DarkSide666:
1) set Query methods as public or protected
2) extend test case class to allow testing of protected/private methods by using Reflection class
3) tried to make composer autoload-dev work without success

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,6 +20,9 @@
         "codeclimate/php-test-reporter": "dev-master"
     },
     "autoload": {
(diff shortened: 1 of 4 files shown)


romaninsh:
Hi. I have updated your pull request. Generally I like the idea of using private methods and being able to test them, but the test-case inheritance creates unnecessary complexity.

I have moved private method invoker into a static class, and re-enabled tests which you have commented.

PR #8, opened 2016-02-09
Feature/implement raw expressions and rendering

romaninsh:
Implemented Expressions.

Unlike the old dsql() where expression was done by same dsql(), here Expression is actually a sub-class of Query, which allows us to nicer split functionality.

Expression would have a template, will know how to parse arguments and how to add nested expressions. Query extends Expressions with some basic templates (select, insert, update etc) and defines logic for methods such as field(), where(), etc.

This pull request moves some of the test-cases into TestExpression() but overall increasing readability.


romaninsh:
I'll need to work a bit more on the documentation for Expression.

PR #20, opened 2016-03-12
Feature/add or support

romaninsh:


src/Expression.php
```diff
@@ -40,7 +40,13 @@ class Expression implements \ArrayAccess {
     public $params=[];
 
 
-    function __construct($template = null, $arguments = null)
+    /**
+     * Specifying options to constructors will override default
+     * attribute values of this class
+     *
+     * @param array $options will initialize class properties
```
DarkSide666: Incorrect description. Should be @param array $template and @param array $arguments

PR #21, opened 2016-03-12
implemented insert and update fully. resolve #15

### LINKED ISSUE #15: Implement INSERT, UPDATE, DELETE basics
Those should now be easily doable. Create implementation and necessary tests.

romaninsh:


src/Query.php
```diff
@@ -519,6 +540,92 @@ public function _render_having()
     }
     // }}}
 
+    // {{{ Set field implementation
+    /**
+     * Sets field value for INSERT or UPDATE statements.
+     *
+     * @param string $field Name of the field
```
DarkSide666: $field can also be array

romaninsh: agree. test case missing also.

romaninsh:
I've restored and merged develop, although it was a bit messy, the "files changed" looks clean now.

I have also realised that I don't have documetation for this PR (or the previous), so i need to add that before merging.

romaninsh:
Added documentation as promised. should be OK to merge, but don't forget to review.

PR #22, opened 2016-03-20
Feature/result set

DarkSide666:


src/Expression.php
```diff
@@ -111,6 +109,27 @@ public function offsetGet($offset) {
     }
 
     /**
+     * Use this instead of "new Expression()" if you want to automatically bind
+     * expression to the same connection as the parent.
+     */
+    public function expr($expr, $options = [])
+    {
+        $options['connection'] = $this->connection;
+        $class = get_class($this);
```
DarkSide666: Unused variable. Probably in next line should be something like `new $class(...)`

src/Expression.php
```diff
@@ -283,4 +302,67 @@ public function getDebugQuery()
...
+                if (!$statement->bindValue($key,$val,$type)) {
+                    throw new Exception('Unable to bind parameter');
+                }
+            }
+
+            $statement->setFetchMode(\PDO::FETCH_ASSOC);
+            $statement->execute();
+            return $statement;
+            
+        } else {
```
DarkSide666: `else` statement here is not needed

src/Expression.php
romaninsh: I was thinking about duplicating same class, but not sure how to approach this properly

DarkSide666: I guess `return new Expression` is OK. At least IDEs and analyzers will like that better :)

PR #40, opened 2016-03-27
Implemented order() and limit()

romaninsh:


src/Query.php
```diff
@@ -700,6 +700,117 @@ protected function _render_set_values()
...
+    // {{{ Limit 
+    /**
+     * Limit how many rows will be returned.
+     *
+     * @param int $cnt   Number of rows to return
+     * @param int $shift Offset, how many rows to skip
+     *
+     * @return DB_dsql $this
+     */
+    public function limit($cnt, $shift = null)
```
DarkSide666: In SQL it's `limit [offset,]row_count`, but in method parameters it's in opposite order - `cnt, shift`. This can be a bit hard to get used to and not mix these parameters.

src/Query.php
```diff
@@ -700,6 +700,117 @@ protected function _render_set_values()
...
+    /**
+     * Orders results by field or Expression. See documentation for full
+     * list of possible arguments.
+     *
+     * $q->order('name');
+     * $q->order('name desc');
+     * $q->order('name desc, id asc')
+     * $q->order('name',true);
+     *
+     * @param string $order Order by
```
DarkSide666: type should be `string|array`

src/Query.php
```diff
@@ -700,6 +700,117 @@ protected function _render_set_values()
...
+     * @param string $desc  true to sort descending
```
DarkSide666: type should be `string|bool`

PR #41, opened 2016-03-28
implement missing methods insert(), update(), replace(), delete()

romaninsh:


tests/db/SelectTest.php
```diff
@@ -68,21 +81,95 @@ public function testBasicQueries()
...
+    {
+        $this->q('employee')->truncate();
+        $this->assertEquals(
+            0,
+            $this->q('employee')->field(new Expression('count(*)'))->getOne()
+        );
+    }
+    */
+
+    public function testOtherQueries()
```
romaninsh: Technically this needs to be in other db/* file because this one is called testSelectQueries, or we can just rename it.

PR #58, opened 2016-04-14
add nice VarDump suppor

romaninsh:
Previously var_dump($query) would output a lot of garbage:

```
class atk4\dsql\Query#382 (11) {
  public $templates =>
  array(6) {
    'select' =>
    string(64) "select [field] [from] [table][join][where][having][order][limit]"
    'delete' =>
    string(36) "delete [from] [table][where][having]"
    'insert' =>
    string(64) "insert into [table_noalias] ([set_fields]) values ([set_values])"
    'replace' =>
    string(65) "replace into [table_noalias] ([set_fields]) values ([set_values])"
    'update' =>
    string(40) "update [table_noalias] set [set] [where]"
    'truncate' =>
    string(30) "truncate table [table_noalias]"
  }
  public $mode =>
  string(6) "select"
  public $defaultField =>
  string(1) "*"
  protected $main_table =>
  string(4) "user"
  protected $template =>
  NULL
..
```

Now this is cleaned up like this:
```
class atk4\dsql\Query#386 (7) {
  public $R =>
  string(20) "select * from `user`"
  public $mode =>
  string(6) "select"
  public $template =>
  NULL
  public $params =>
  array(0) {
  }
  public $connection =>
  NULL
  public $main_table =>
  string(4) "user"
  public $args =>
  array(2) {
    'custom' =>
    array(0) {
    }
    'table' =>
    array(1) {
      [0] =>
      array(2) {
        ...
      }
    }
  }
}
```

for expressions it's also cleaner:

```
class atk4\dsql\Expression#388 (5) {
  public $R =>
  string(31) "Expression could not render tag"
  public $template =>
  string(13) "Hello [world]"
  public $params =>
  array(0) {
  }
  public $connection =>
  NULL
  public $args =>
  array(1) {
    'custom' =>
    array(0) {
    }
  }
}
```

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -383,6 +383,26 @@ class Expression implements \ArrayAccess, \IteratorAggregate
         return $d.' <span style="color:gray">[' . implode(', ', $pp) . ']</span>';
     }
 
+    function __debugInfo()
+    {
+
+        $arr = [
+            'R'=>false,
+            'template'=>$this->template,
+            'params'=>$this->params,
+            'connection'=>$this->connection,
+            'args'=>$this->args,
+        ];
+
+        try {
+            $arr['R'] = $this->render();
+        } catch (Exception $e) {
+            $arr['R'] = $e->getMessage();
+        }
+
+        return $arr;
+    }
+
     /**
      * Execute expression
      *
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -396,7 +396,7 @@ class Expression implements \ArrayAccess, \IteratorAggregate
 
         try {
             $arr['R'] = $this->render();
-        } catch (Exception $e) {
+        } catch (\Exception $e) {
             $arr['R'] = $e->getMessage();
         }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: php
 php:
     - '5.5'
+    - '5.6'
+    - '5.7'
     - '7.0'
 
 services:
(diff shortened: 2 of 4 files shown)


src/Expression.php
```diff
@@ -383,6 +383,26 @@ public function getDebugQuery()
...
+            'R'=>false,
+            'template'=>$this->template,
+            'params'=>$this->params,
+            'connection'=>$this->connection,
+            'args'=>$this->args,
+        ];
+
+        try {
+            $arr['R'] = $this->render();
+        } catch (Exception $e) {
```
DarkSide666: should be `\Exception`

src/Expression.php
```diff
@@ -383,6 +383,26 @@ public function getDebugQuery()
...
+            'R'=>false,
+            'template'=>$this->template,
+            'params'=>$this->params,
+            'connection'=>$this->connection,
+            'args'=>$this->args,
+        ];
+
+        try {
+            $arr['R'] = $this->render();
+        } catch (\Exception $e) {
```
DarkSide666: even better would be:
```
catch (Exception $e) {
    $arr['R'] = $e->getMessage();
    $arr['R-params'] = $e->getParams();
} catch (\Exception $e) {
    $arr['R'] = $e->getMessage();
}
```
to show all params too in case this is dsql exception :)

romaninsh: hehe. ok. but leave as-is. we do not know where that information will end up. It's unsafe to expose Params.

PR #59, opened 2016-04-14
docs review

DarkSide666:


docs/queries.rst
```diff
@@ -21,7 +21,7 @@ expression or you can execute it in exchange for result set.
 
 Quick Example::
 
-    use atk4\dsql;
+    use atk4\dsql\Query;
```
romaninsh: i think we should rewrite this using connection. I can review later.

PR #64, opened 2016-04-14
Feature/refactor table

DarkSide666:
* now you can add as many sub-queries as tables as you want to you query
* alias is mandatory if you set table as object (expression, query or whatever)
* main_table functionality almost not changed only little bit tweaked
* few psr fixes
* more tests

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -40,7 +40,7 @@ class Query extends Expression
     public $defaultField = '*';
 
     /**
-     * Name of base table to use when using default join().
+     * Name or alias of base table to use when using default join().
      *
      * This is set by table(). If you are using multiple tables,
      * then $main_table is set to false as it is irrelevant.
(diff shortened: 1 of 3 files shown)


src/Query.php
```diff
@@ -240,29 +234,37 @@ public function table($table, $alias = null)
...
+        // process tables one by one
+        foreach ($this->args['table'] as $alias => $table) {
 
-            // This will wrap into () if Query object is used here
-            return $this->_consume($this->args['table']);
-        }
+            // throw exception if we don't want to add alias and table is defined as Expression
+            if ($add_alias === false && $table instanceof Expression) {
+                throw new Exception('Table cannot be expression for UPDATE, INSERT etc. queries');
+            }
```
romaninsh: indentation issue?

src/Query.php
```diff
@@ -240,29 +234,37 @@ public function table($table, $alias = null)
      * Renders part of the template: [table]
      * Do not call directly.
      *
+     * @param bool $add_alias Should we add aliases, see _render_table_noalias()
+     *
      * @return string Parsed template chunk
      */
-    protected function _render_table()
+    protected function _render_table($add_alias = true)
```
romaninsh: wasn't there a recomendation to keep default values "null" or "false" ?

PR #73, opened 2016-04-15
add support for {tag} in expressions

romaninsh:
## Added {tag} support:
```
new Expression('hello {tag}', ['world']);
```

now generates:

```
hello `world`
```

## refactored _escape

_escape now very strictly escapes it's argument (or array) by adding back-tick.

## added _escapeSoft

Emulates how old _escape worked but also added support for changing 
```
user.name
```
into
```
`user`.`name`
```

You can also call _consume($arg, 'soft-escape');

## removed second argument from field()

Now field always uses ($what, $alias). If you nee to specify table simply use field('table.field');

## refactored table() and it's rendering a bit

Mainly to allow use expression as a table without punishment (queries should still be aliased)

## Updated documentation.

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -123,7 +123,8 @@ class Query extends Expression
 
         // recursively add array fields
         if (is_array($field)) {
-            if (!is_null($alias)) {
+
+            if ($alias !== null) {
                 throw new Exception([
                     'Alias must not be specified when $field is an array', 
                     'alias'=>$alias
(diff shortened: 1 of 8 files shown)


src/Query.php
```diff
@@ -83,60 +83,80 @@ class Query extends Expression
...
+     * Associative array will assume that "key" holds the field alias.
+     * Value may be field name, Expression or Query.
+     *  $q->field(['alias' => 'name', 'alias2' => 'mother.surname']);
+     *  $q->field(['alias' => $q->expr(..), 'alias2' => $q->dsql()->.. ]);
      *
-     * You can use $q->dsql() for subqueries. In such case field alias is mandatory
-     *  $q->field( $q->dsql()->table('x')... , 'alias');    // must always use alias
+     * If you need to use funky name for the field (e.g, one condaiting 
+     * a dot or space), you should wrap it into expression:
+     *  $q->field($q->expr('{}', ['fun...ky.field']), 'f');
```
DarkSide666: What this will generate as result? I guess ``fun..ky.field``, right?

romaninsh: yes, a weirdly named field, still usable.

PR #75, opened 2016-04-27
Feature/support field names with spaces

DarkSide666:
* group by and order by now uses soft-escape
* added few test cases for that
* order by will no more try to explode string by dot, soft-escape does that in rendering phase
* order by tries to be smart when exploding field name by space and searching for ordering keyword. It'll only explode if correct ordering keyword is actually found at the end of string.
* added test cases for group by and order by with strange field names.

resolves #64

### DIFF
--- a/docs/expressions.rst
+++ b/docs/expressions.rst
@@ -300,6 +300,21 @@ circumstances.
       $query->_consume('first_name');  // `first_name`
       $query->_consume($other_query);  // will merge parameters and return string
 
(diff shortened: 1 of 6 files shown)


romaninsh:
Order by array also works? Soft or hard escape?

romaninsh:
What do you think we make escapeSoft call arraymap with hard escape?

romaninsh:
I have added few improvements, dark can you check?

PR #82, opened 2016-06-19
Feature/fix cc issues 20160620

### LINKED ISSUE #81: Improve code climate
https://codeclimate.com/github/atk4/dsql/src/Query.php

Need to improve issues in this file.

DarkSide666:
fixes #81 

https://codeclimate.com/github/atk4/dsql/compare/feature/fix-cc-issues-20160620

### DIFF
--- a/docs/queries.rst
+++ b/docs/queries.rst
@@ -719,6 +719,12 @@ Other Methods
     $q->select(); // select calc_found_rows `name` from `test`
     $q->insert(); // insert ignore into `test` (`name`) values (`name` = 'John')
 
+.. php:method:: _set_args($what, $alias, $value)
+
+    Internal method which sets value in :php:attr:`Expression::args` array.
+    It doesn't allow duplicate aliases and throws Exception in such case.
+    Argument $what can be 'table' or 'field'.
+
 
 Properties
 ==========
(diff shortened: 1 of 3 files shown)


DarkSide666:
Method `_set_args` can also be written for not limited depth something like this:

```
/**
 * @param array $path Like ['table', $alias] or ['where', null]
 * @param mixed $value
 * @param bool $deny_overwrite
 */
    protected function _set_args($path, $value, $deny_overwrite = true)
    {
        $new_key = array_pop($path);
        
        // navigate
        $a = &$this->args;
        foreach ($path as $key) {
            if (!is_array($a)) {
                $a = array();
            }
            $a = &$a[$key];
        }

        // set value
        if ($new_key === null) {
            $a[] = $value;
        } else {
            if ($deny_overwrite && isset($a[$new_key])) {
                throw new Exception([
                    'Args element already set',
                    'path' => implode('/', $path)
                ]);
            }
            $a[$new_key] = $value;
        }
}
```

and used in all class methods, but I guess that will make performance worse because of additional method call.

romaninsh:
Add documentation mention

DarkSide666:
will do probably tonight

romaninsh:
i think just mentioning this under internal method with some brief paragraph is enough.

DarkSide666:
simple method description added in docs

PR #83, opened 2016-06-22
Improve test coverage, fix few small bugs

DarkSide666:


romaninsh:
Dark, I've released 1.0.1 which is now our STABLE DSQL version. I'm merging your PR which will target at 1.0.2 now.

tests/ExpressionTest.php
```diff
@@ -258,6 +258,26 @@ public function testNestedExpressions()
     }
 
     /**
+     * @covers ::__toString
+     * @expectedException \LogicException
+     */
+    public function testToStringException1()
```
romaninsh: this triggers crash in test-suite, but because it produced "zero" output it wasn't failed.

PR #100, opened 2016-07-21
Allow set field to be an expression

romaninsh:
Allow field to be expression in set()

### DIFF
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -1367,7 +1367,7 @@ class QueryTest extends \PHPUnit_Framework_TestCase
     }
 
     /**
-     * Field name can be expression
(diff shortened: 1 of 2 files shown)


tests/QueryTest.php
```diff
@@ -1367,10 +1367,9 @@ public function testSetException1()
     }
 
     /**
-     * Field name should be array or string.
+     * Field name can be expression.
      *
      * @covers ::set
-     * @expectedException Exception
      */
     public function testSetException2()
```
DarkSide666: probably different name for this method would be good :)

romaninsh: haha... no complains from codeclimate...

PR #106, opened 2016-08-04
Use atk core exception and add more info in exceptions

DarkSide666:
* will use \atk4\core\Exception everywhere
* added try/catch in Exception->execute which will add additional info in exception - query itself :)
* no more variables in exception messages - good for localization in future
* add more params in exceptions - help in debugging
* expectedExceptionMessage should not be used in test because exception texts contain params too and can be changed

### DIFF
--- a/tests/db/TransactionTest.php
+++ b/tests/db/TransactionTest.php
@@ -152,7 +152,7 @@ class TransactionTest extends \PHPUnit_Extensions_Database_TestCase
                     ->set(['id' => 4, 'FOO' => 'bar', 'name' => 'Jane', 'surname' => 'Doe', 'retired' => 0])
                     ->insert();
             });
-        } catch (\PDOException $e) {
+        } catch (\Exception $e) {
             // ignore
         }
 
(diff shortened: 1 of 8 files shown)


tests/db/TransactionTest.php
```diff
@@ -152,7 +152,7 @@ public function testTransactions()
                     ->set(['id' => 4, 'FOO' => 'bar', 'name' => 'Jane', 'surname' => 'Doe', 'retired' => 0])
                     ->insert();
             });
-        } catch (\PDOException $e) {
+        } catch (\Exception $e) {
```
romaninsh: i think PDO i more appropriate here.

romaninsh: oh it's test script so we dont care ;)

DarkSide666: yeah it's test script and need to have Exception instead of PDOException because now Expresion->execute catches PDOException and throws its  own exception insted with some additional info.

romaninsh: ah i see.

PR #123, opened 2017-12-30
Allow fields to be in an SQL query

romaninsh:
There is a bit of unclarity with the implementation, so this is a hack to workaround some limitations. This is to be refactored when second major persistence is added to Agile Data, but here is the reasoning.

1. Currently Model Field must implement Expressionable in order to query data from SQL persistence.
2. There no such limitation on Array persistence, any field will be loaded
3. DSQL shouldn't rely on Agile Data. Agile Data field shouldn't rely on DSQL.

So the proper solution here would be to move expressionable into Agile Core and require all model fields to implement it and use it regardless of the database, although it's not realistic until we have a better understanding of how multiple complex (with expressions) persistences co-exist.

The work-around here simply replaces what would be an exception with an acceptable approach field embedding.


romaninsh:
never mind. found a better way.

PR #124, opened 2018-01-04
Feature/oracle support

romaninsh:
Adds support for Oracle database.


src/Connection.php
```diff
@@ -68,13 +68,21 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+                    'connection'       => new \PDO($dsn, $user, $password),
+                    'expression_class' => 'atk4\dsql\Expression_MySQL',
+                    'query_class'      => 'atk4\dsql\Query_MySQL',
                 ], $args));
             case 'sqlite':
                 return new self(array_merge([
-                    'connection'  => new \PDO($dsn, $user, $password),
-                    'query_class' => 'atk4\dsql\Query_SQLite',
+                    'connection'       => new \PDO($dsn, $user, $password),
+                    'expression_class' => 'atk4\dsql\Expression',
```
romaninsh: technically - not needed. default.

src/Connection.php
```diff
@@ -68,13 +68,21 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+                    'query_class'      => 'atk4\dsql\Query_SQLite',
+                ], $args));
+            case 'oci':
+                return new self(array_merge([
+                    'connection'       => new \PDO($dsn, $user, $password),
+                    'expression_class' => 'atk4\dsql\Expression',
```
romaninsh: same.

src/Query_Oracle_Abstract.php
```diff
@@ -0,0 +1,15 @@
...
+namespace atk4\dsql;
+
+/**
+ * Perform query operation on Oracle server.
+ *
+ * @license MIT
+ * @copyright Agile Toolkit (c) http://agiletoolkit.org/
+ */
+abstract class Query_Oracle_Abstract extends Query
+{
```
romaninsh: please get rid of this class..

DarkSide666: leaving it for now because i have feeling that something could go there - Oracle specific feature which all Oracle versions share.

PR #126, opened 2018-01-20
Feature/oracle lastid

romaninsh:
YUK, so there is this problem with Oracle:

SQLSTATE[IM001]: Driver does not support this function: driver does not support lastInsertId()

Agile Data can use beforeSave / beforeInsert hook to get sequence number and use it inside insert, but we need ability to specify custom lastInsertID() on DSQL level.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -298,7 +298,7 @@ class Connection
 
     public $override_last_insert_id = null;
 
-    function lastInsertID()
+    public function lastInsertID()
     {
         if ($this->override_last_insert_id) {
             $id = $this->override_last_insert_id;
@@ -306,6 +306,7 @@ class Connection
 
             return $id;
         }
+
         return $this->connection->lastInsertID();
     }
 }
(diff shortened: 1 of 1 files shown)


src/Connection.php
```diff
@@ -295,4 +295,18 @@ public function rollBack()
 
         return false;
     }
+
+    public $override_last_insert_id = null;
```
DarkSide666: How about ability to use callable function here and/or nested DSQL expression (sub-select) like `select xxx_seq.nextval from dual` ?

romaninsh: I put this patch quickly together, probably there is a better implementation.

After this I started to wonder, why couldn't ID be passed in as a trigger?

And if during "insert" we somehow specify "ID" value, perhaps we shouldn't even attempt to get last_insert_id and simply set it to value of data[id_field]

romaninsh:
This implementation is not ideal, so I propose to close this and implement this differently.

DarkSide666:
closing

Issue #129, opened 2018-01-29
[epic] Add support for PostgreSQL


romaninsh:
For a proper implementation of PostgreSQL:

 - [ ] add class and detection #127
 - [x] id_field #128 
 - [x] add test-suite #27

PR #130, opened 2018-01-29
Create a driver for postgresql

gartner:
Create a Query_PgSQL-class and make dsql use this if a postgresql database dsn is supplied to \atk4\dsql\Connect::connect()

It works, when I use it with the CRUD-component from atk4\ui, but I will need some guidance to write tests for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ cache:
   directories:
     - $HOME/.composer/cache
 
+addons:
+  postgresql: "9.3"
+
 services:
     - mysql
     - postgresql
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_script:
     - composer self-update
     - composer install --no-ansi
     - mysql -e 'create database dsql_test;'
-    - psql -c 'create database atk4-dsql-test;' -U postgres
+    - psql -c 'create database "atk4-dsql-test";' -U postgres
     - mkdir -p build/logs
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,17 +12,20 @@ cache:
 
 services:
     - mysql
+    - postgresql
 
 before_script:
     - composer self-update
     - composer install --no-ansi
     - mysql -e 'create database dsql_test;'
+    - psql -c 'create database atk4-dsql-test;' -U postgres
     - mkdir -p build/logs
 
 script:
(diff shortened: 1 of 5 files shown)


romaninsh:
This looks like a great PR, but still problems in travis:

`psql -c 'create database atk4-dsql-test;' -U postgres`

gartner:
I'm running the same tests that exists for mysql. But some of them are failing. But not because of the code, but because of the database. 
There are this one:
´´´$this->q()->field(new Expression('[]+[]', [3, 3]), 'now')->get()´´´
This is actually both creating and running a correct, valid query: ´´´select 3+3 "now"´´´ but when running it, postgresql complains that it cannot find a usefull operator for 'unknown + unknown'. This seems to be caused by using prepare() when creating the query: ´´´:a+:b "now"´´´and then passing these parameters when executing the query. Postgresql will  try to guess the operator for + when preparing, but at that time, it still does not know the type of the parameters. 
If I run a type-hinted version:
´´´$this->q()->field(new Expression('[]::int+[]::int', [3, 3]), 'now')->get()´´´ all is good. (The ::int will cast these values to integers.)
But I do not think that syntax is valid for mySql?

A later query is testing with the function IFNULL() - works on mysql, but IFNULL() does not exist in postgresql. So it fail.

So, some of the tests are failing, but not as a result of the php code, but the backend database being different.

How should this be solved in the tests? dsql being some kind of database abstraction layer, I would thing that the same testsuite should be able to run on different databases. So, should we rewrite the queries that the tests run, to work on all the supported databases? Should the testsuite use if()'s and only run some of these tests on the databases that they work on?

romaninsh:
@gartner thanks again for all the work. Inspired by your changes I have installed PostgreSQL locally and trying to get things moving with your branch. 

A proper way to handle IFNULL should be through checking persistence type (you can do $connection instanceof postgres) and then skipping  the test#:

```
$this->markTestSkipped('not supported on postgres');
```

you can add alternative test and skip it for vendors other then postgres.


For the type-casting, the fault could be in PDO::PARAM_INT, where it does not bind parameter as integer in the first place. 

https://github.com/atk4/dsql/blob/b1241953d8f6fb7fe7d0b764a4f4815b48e9ba39/src/Expression.php#L556

If there is no way to bind integer value, you can use "cast([] as decimal)" perhaps, I'm not sure if that would work on both databases, but worth a try.

Finally, you can try duplicating / skipping test if nothing else works.

romaninsh:
Implements #129

romaninsh:
Style will be auto-corrected after merge. Looks like all good to merge.

PR #131, opened 2018-01-30
Fix: SQL does not guarantee the order of retreived rows unless explic…

gartner:
The tests are comparing an array with the data retreived from database. Phpunit will fail if the two arrays are not in the same order.
SQL does not guarantee the order of rows retrieved, unless the 'ORDER BY' is added to the query. 
This pull-request adds the order()-clause to queries.


romaninsh:
Nice and simple fix, thanks!

Issue #134, opened 2018-02-07
Passing a \PDO to connect() assumes mysql


gartner:
It is possible to pass a \PDO-instance to Connect::connect() instead of a dsn. 

But when doing so, it is assumed to be a connection to a mysql database.
I have a fix here: https://github.com/gartner/dsql/tree/fix-pdo - but it seems a bit messy. Especially the tests, which although they pass, I am not satisfied with. 
Part of the messy-ness of the tests is that phpunit would not use dataProviders when run on my machine. 

So, now the question is: should I create a pull request, or perhaps instead dsql should _not_ allow a \PDO to be passed to connect() ?

romaninsh:
you should pass it to `new Connection_PostgreSQL($dsn)`.

DarkSide666:
Yes PR would be good, but I think it should work as Romans said above.
If there will be PR, then please also add me as reviewer for it because I have suffered with these PDO issues to while testing with Oracle.

PR #135, opened 2018-02-08
Fix pdo

gartner:
Connection::connect() can take an instance of PDO instead of a dsn. 
But when passing a PDO-instance, it would assume it to be to a mysql database. This could fail, now that other databases are also supported. See #134 

This will fix it, see what kind of database the PDO is connected to, and setup Query- and Expression-classes accordingly.

In future version, when a BC-break can be made, support for passing a PDO to connect() should probably be dropped...

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -60,7 +60,7 @@ class Connection
         if ($dsn instanceof \PDO) {
             $driver = $dsn->getAttribute(\PDO::ATTR_DRIVER_NAME);
             $connectionClass = '\\atk4\\dsql\\Connection';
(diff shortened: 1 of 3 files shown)


DarkSide666:
please also fix styleci errors: https://styleci.io/analyses/8j90r7

DarkSide666:
Looks OK. At some point we will probably need to refactor PDO/Connection/Persistence a bit because now they all 3 partly duplicate code.

romaninsh:
Many thanks Dark and Gartner, great work!

Issue #74, opened 2016-04-17
Add support for CASE expression


romaninsh:
Currently in order to include CASE expression in your field list, one would have to write:

```
$query->field(new Expression("case {} when [] then {} when [] then {} else null end", [
    'status',
    'New', 't2.expose_new'
    'Used', 't2.expose_used'
]), 'expose');
```

http://stackoverflow.com/questions/8600671/mysql-select-statement-with-case-or-if-elseif-not-sure-how-to-get-the-result

Similarly how we are tackling orExpr() we can also add support for caseExpr() which should have a more pleasant format:

```
$query->field($query->caseExpr('status')
  ->when('New', 't2.expose_new')
  ->when('Used', 't2.expose_used')
  ->else(null), 'expose');
```

DarkSide666:
We should support 2 types of CASE expressions:

**Normal form:**
```
CASE WHEN <condition> THEN <result>
    WHEN <condition> THEN <result>
    [WHEN <condition> THEN <result>
     ...]
    [ELSE <result>]
END
```

```
$query->field($query->caseExpr()
  ->when(['status','New'], 't2.expose_new')
  ->when(['status', 'like', '%Used%'], 't2.expose_used')
  ->else(null), 'expose');

CASE
    WHEN "status"='New' THEN "t2"."expose_new"
    WHEN "status" like '%Used%' THEN "t2"."expose_used"
    ELSE null
END "expose"
```

**And simple form:**
```
CASE <common operand>
     WHEN <expression> THEN <result>
    [WHEN <expression> THEN <result>
     ...]
    [ELSE <result>]
END
```

```
$query->field($query->caseExpr('status')
  ->when('New', 't2.expose_new')
  ->when('Used', 't2.expose_used')
  ->else(null), 'expose');

CASE "status"
    WHEN 'New' THEN "t2"."expose_new"
    WHEN 'Used' THEN "t2"."expose_used"
    ELSE null
END "expose"
```

PR #138, opened 2018-03-12
Implement CASE WHEN/THEN

DarkSide666:
fixes https://github.com/atk4/dsql/issues/74


DarkSide666:
PHP 5.6 don't like method name `else`. PHP 7 is fine with it.
If we have to support PHP 5.6 then need to change that method name to something like `elseCase` :(
Decision is yours :)
https://travis-ci.org/atk4/dsql/jobs/352135521#L615

PR #139, opened 2018-03-19
Implement normalizeDSN()

DarkSide666:
* implement normalizeDSN()
* Connection now use normalizeDSN()
* test coverage

P.S. normalizeDSN() will also be used in \atk4\data\Persistence to avoid code duplication between repos.

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -45,6 +45,43 @@ class Connection
     }
 
     /**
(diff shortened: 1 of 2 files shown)


romaninsh:
Imants while you are at it, can you make somehow possible to detect inside Persistence what database type it is? e.g. $persistence_sql->sql_vendor ?

PR #142, opened 2018-04-17
add resource support

DarkSide666:
we need this for clob/blob/resource support


DarkSide666:
It works for reading clob/blob from database. We don't have tests for Oracle.
Anyway this one additional condition will not make any bad anyway :)

romaninsh:
i suppose. codecov is not integrated too.

PR #143, opened 2018-04-19
replace bindValue to bindParam and use trick for LOBs

DarkSide666:


DarkSide666:
I hope so too. Maybe that will have small impact on memory consumption because it duplicates parameters, but in one query you don't have 1000 parameters so i guess it should be fine.

PR #154, opened 2019-01-09
Feature/implement action interface

romaninsh:
Implements ResultSet interface.

would be useful for https://github.com/atk4/data/pull/377


DarkSide666:
Need to at least describe each method what it does and why it is needed.

romaninsh:
It is an interface. But I'll describe.

PR #161, opened 2019-08-24
Show Error Message in Exceptions Message 

PhilippGrashoff:
If this exception is output by some non-atk output which uses functions like getColorfulText(), the only thing displayed is `DSQL got Exception when executing this query` Thats quite annoying as you have do not get any hint whats wrong.

A good example is phpunit which outputs the Exceptions message. While `DSQL got Exception when executing this query` really is not helpful, `DSQL got Exception when executing this query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'ident' in 'field list'` is.

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -606,7 +606,7 @@ class Expression implements \ArrayAccess, \IteratorAggregate, ResultSet
                 $statement->execute();
             } catch (\Exception $e) {
                 $new = new Exception([
-                    'DSQL got Exception when executing this query',
+                    'DSQL got Exception when executing this query: '.$e->getMessage(),
                     'error' => $e->getMessage(),
                     'query' => $this->getDebugQuery(),
                 ]);


DarkSide666 (approved):
On one hand it's good change because helps you to see what's going wrong (i have made similar change in my extended code somewhere too). But on other hand - maybe we should show additional parameters (like 'error' here) in getColorfulText() instead?

PhilippGrashoff:
Hi, my inital text was not really clear: I meant "non-atk functions that do NOT use functions like getColorfulText()"
GetColorfulText() rendered that exception nicely before. But any non-atk Exception catch would use the exceptions message, which, in this case, is not helpful at all.

DarkSide666:
Ah I see. Yes you're right, but still I would like to see Romans decision on this. I think it's fine.

romaninsh:
Few reasons why the query is not included in the message:
 - localization
 - injecting all sorts of nastiness

If this can be implemented using a named parameter, then it would be acceptable. @abbadon1334 - topic to discuss at next meeting?

romaninsh (changes requested):
Would rather have it implemented through localized parameter.

abbadon1334:
Yes, now I'm travelling back to Dracula's land, tomorrow I will be at the meeting. I want to speak even on addSolution, I think this can be a good starting point for localization.

mvorisek:
Closing as this needs to be solved on the printer side to display the params. The edited code does not even use the full message now - it sets `error` to `$e->errorInfo[2]` text from the `\PDOException`.

PR #164, opened 2020-01-05
handle empty array value gracefully

georgehristov:
Fixes https://github.com/atk4/data/issues/496


georgehristov:
Thank you @DarkSide666

PR #166, opened 2020-02-07
Fix CS to PSR-12

mvorisek:
Add CS config and fix PSR-12 errors.


mvorisek:
Closing in favor of PR #190

PR #170, opened 2020-03-09
implements template tags `{{...}}` for soft escape

DarkSide666:


DarkSide666:
added short reference to template tags in docs

PR #171, opened 2020-03-22
WITH implementation

### LINKED ISSUE #117: Add support for WITH clause
Add support in the `Query` class to implement [https://en.wikipedia.org/wiki/Hierarchical_and_recursive_queries_in_SQL#Common_table_expression](CTE) using SQL `WITH` clause.

The signature of the function could be the following:

``` php
public function with($expr, $table_alias, $fields = [])
public function withRecursive($expr, $table_alias, $fields = [])
```

The alias in this case should be mandatory (single identifier or array), while expr could be any kind of subtable expression, or even a union. $fields is optional.

To produce the following query

``` sql
WITH RECURSIVE
  cnt(x) AS (VALUES(1) UNION ALL SELECT x+1 FROM cnt WHERE x<1000000)
SELECT x FROM cnt;
```

you would do:

``` php
$expr = newExpression('values(1) union all select x+1 from cnt where x<1000000');

$q = new Query();
$q->table('cnt');
$q->field('x');
$q->withRecursive($expr, 'cnt', ['x']);
```

It should be possible to execute with() multiple times.

DarkSide666:
This PR implements `WITH` statement.
fix #117 

```
$quotes = $q->table('quotes')
    ->field('emp_id')
    ->field($q->expr('sum([])', ['total_net']))
    ->group('emp_id');
$invoices = $q()->table('invoices')
    ->field('emp_id')
    ->field($q->expr('sum([])', ['total_net']))
    ->group('emp_id');
$employees = $q
    ->with($quotes, 'q', ['emp','quoted'])
    ->with($invoices, 'i', ['emp','invoiced'])
    ->table('employees')
    ->join('q.emp')
    ->join('i.emp')
    ->field(['name', 'salary', 'q.quoted', 'i.invoiced']);
```

This generates SQL below:
```
with
    `q` (`emp`,`quoted`) as (select `emp_id`,sum(`total_net`) from `quotes` group by `emp_id`),
    `i` (`emp`,`invoiced`) as (select `emp_id`,sum(`total_net`) from `invoices` group by `emp_id`)
select `name`,`salary`,`q`.`quoted`,`i`.`invoiced`
from `employees`
    left join `q` on `q`.`emp` = `employees`.`id`
    left join `i` on `i`.`emp` = `employees`.`id`
```

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -402,7 +402,7 @@ class Query extends Expression
             $s = $this->_escape($alias).' ';
 
             // set cursor fields
-            if ($fields) {
+            if ($fields !== null) {
                 $s .= '('.implode(',', array_map([$this, '_escape'], $fields)).') ';
             }
 
--- a/src/Query.php
+++ b/src/Query.php
@@ -354,29 +354,28 @@ class Query extends Expression
      *
      * @return $this
      */
-    public function with(Query $cursor, string $alias, ?array $fields = null, bool $recursive = false)
+    public function with(self $cursor, string $alias, ?array $fields = null, bool $recursive = false)
     {
         // save cursor in args
         $this->_set_args('with', $alias, [
-            'cursor' => $cursor,
-            'fields' => $fields,
+            'cursor'    => $cursor,
+            'fields'    => $fields,
             'recursive' => $recursive,
         ]);
 
-
         return $this;
     }
-    
+
     /**
      * Recursive WITH query.
      *
-     * @param Query|array $cursor    Specifies cursor query or array [alias=>query] for adding multiple
-     * @param string      $alias     Specify alias for this cursor
-     * @param array  $fields    Optional array of field names used in cursor
+     * @param Query|array $cursor Specifies cursor query or array [alias=>query] for adding multiple
+     * @param string      $alias  Specify alias for this cursor
+     * @param array       $fields Optional array of field names used in cursor
      *
      * @return $this
      */
-    public function withRecursive(Query $cursor, string $alias, ?array $fields = null)
+    public function withRecursive(self $cursor, string $alias, ?array $fields = null)
     {
         return $this->with($cursor, $alias, $fields, true);
     }
@@ -391,7 +390,7 @@ class Query extends Expression
     {
         // will be joined for output
         $ret = [];
-        
+
         if (empty($this->args['with'])) {
             return '';
         }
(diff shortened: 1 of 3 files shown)


src/Query.php
```diff
@@ -345,6 +342,84 @@ protected function _render_from()
...
+        }
+
+        // process each defined cursor
+        $isRecursive = false;
+        foreach ($this->args['with'] as $alias => ['cursor'=>$cursor, 'fields'=>$fields, 'recursive'=>$recursive]) {
+            // cursor alias cannot be expression, so simply escape it
+            $s = $this->_escape($alias).' ';
+
+            // set cursor fields
+            if ($fields) {
```
mvorisek: Test for `$fields === null`.

src/Query.php
```diff
@@ -345,6 +342,84 @@ protected function _render_from()
...
+                $s .= '('.implode(',', array_map([$this, '_escape'], $fields)).') ';
+            }
+
+            // will parameterize the value and escape if necessary
+            $s .= 'as '.$this->_consume($cursor, 'soft-escape');
```
mvorisek: `AS` in UC would be nicer

src/Query.php
```diff
@@ -345,6 +342,84 @@ protected function _render_from()
...
+
+            // is at least one recursive ?
+            $isRecursive = $isRecursive || $recursive;
+
+            $ret[] = $s;
+        }
+
+        return 'with '.($isRecursive ? 'recursive ' : '').implode(',', $ret).' ';
```
mvorisek: Same here - `WITH` in UC

src/Query.php
```diff
@@ -36,7 +36,7 @@ class Query extends Expression
      *
      * @var string
      */
-    protected $template_select = 'select[option] [field] [from] [table][join][where][group][having][order][limit]';
+    protected $template_select = '[with]select[option] [field] [from] [table][join][where][group][having][order][limit]';
```
mvorisek: Does spaces matter here? Somewhere are, somewhere not, why?

src/Query.php
romaninsh: we don't use UC in SQL. I find this "standard" old and silly. I understand not all computers have lower case characters, but we don't see many of them anymore :D

src/Query.php
DarkSide666: done

src/Query.php
DarkSide666: I like UC too, but made it lowercase for consistency. See - all templates currently use lowercase. If we update it to UC, then should do that in all templates not just this one.

src/Query.php
DarkSide666: No space here because in case `with` is not set then we don't want leading space to show up before `select`.
Basically spaces in template are if it's mandatory to fill particular template tags. Otherwise space is added in rendering process if needed. But I agree - that's not very nice if sometimes we do this and sometimes different.

PR #172, opened 2020-03-28
Fix reference to property unset

mvorisek:
I think the 2nd reference to property should be unset too.


mvorisek:
Based on the tests probably not.

PR #174, opened 2020-03-31
Improve DSQL exception for SQL

mvorisek:
Before:
image

After:
image

Now the SQL execute exception has it's own class, so it can be better tested and error code/message is now much better accessible.

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -615,12 +615,19 @@ class Expression implements \ArrayAccess, \IteratorAggregate, ResultSet
             try {
                 $statement->execute();
             } catch (\Exception $e) {
-                $new = new Exception([
-                    'DSQL got Exception when executing this query',
-                    'error' => $e->getMessage(),
-                    'query' => $this->getDebugQuery(),
-                ]);
-                $new->by_exception = $e;
+                $msg = 'DSQL got Exception when executing this query';
+                if ($e instanceof \PDOException) {
+                    $new = new Exception_SQL([
+                        $msg,
+                        'error' => $e->errorInfo[2],
+                        'query' => $this->getDebugQuery(),
+                    ], $e->errorInfo[1]);
+                } else { // code should never get here with PDO driver
+                    $new = new Exception_SQL([
+                        $msg,
+                        'query' => $this->getDebugQuery(),
+                    ], null, $e);
+                }
 
                 throw $new;
             }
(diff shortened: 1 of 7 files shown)


mvorisek:
Before merging please add 1 test quering `SELECT not_existing_column` which should throw the new `Exception_SQL`.

src/Expression.php
```diff
@@ -614,13 +614,12 @@ public function execute($connection = null)
...
-                $new = new Exception([
+            } catch (\PDOException $e) {
+                $new = new ExecuteException([
                     'DSQL got Exception when executing this query',
-                    'error' => $e->getMessage(),
+                    'error' => $e->errorInfo[2],
                     'query' => $this->getDebugQuery(),
-                ]);
-                $new->by_exception = $e;
+                ], $e->errorInfo[1]);
```
mvorisek, line 622: @DarkSide666 I was not able to find any test testing that the SQL query/execution may fail.

Can you please add 1 test quering `SELECT non_existing_column` which should throw the new `ExecuteException` with `1054` code (`Unknown column` code) at the appropriate place? Thanks.

mvorisek:
I was not able to find any test testing that the SQL query/execution may fail.

Can someone please add 1 test quering `SELECT non_existing_column` which should throw the new `ExecuteException` with `1054` code (`Unknown column` code) at the appropriate place?

DarkSide666 (approved):
LGTM

DarkSide666:
@mvorisek Please see test i added. It will throw PDO exception. Not sure what you mean by adding test.

tests/ExceptionTest.php
```diff
@@ -2,13 +2,20 @@
...
+use atk4\dsql\Connection;
 use atk4\dsql\Expression;
+use atk4\dsql\Query;
 
 /**
  * @coversDefaultClass \atk4\dsql\Exception
  */
 class ExceptionTest extends \atk4\core\PHPUnit_AgileTestCase
 {
+    public function q()
```
mvorisek, line 14: What is the purpose of this?

DarkSide666: Oh yes, can be removed. I added it but after changed test method and it's no more used.

mvorisek: > 
> 
> @mvorisek Please see test i added. It will throw PDO exception. Not sure what you mean by adding test.

I meant https://github.com/atk4/dsql/pull/179/files#diff-78123c449aa133c903d31fc8b34fb882R61 , but it needs to be tested against MySQL/Maria for that specific error code.

PR #175, opened 2020-04-01
Convert scalar class names to ::class

mvorisek:


mvorisek:
`Segmentation fault (core dumped)` in PHP 7.3 tests, how to fix this?

PR #179, opened 2020-04-08
Use ::class instead of string class names

mvorisek:
Fixes #174 
Fixes #180


tests/OracleTest.php
```diff
@@ -33,8 +33,8 @@ public function connect($ver = '')
     {
         return new \atk4\dsql\Connection(array_merge([
             'connection'       => new \PDO('sqlite::memory:'),
-            'query_class'      => 'atk4\dsql\Query_Oracle'.$ver,
-            'expression_class' => 'atk4\dsql\Expression_Oracle',
```
mvorisek, line 37: `\atk4\dsql\Expression_Oracle` class seems to be even not presented

DarkSide666: Should remove that expression_class line (will use default Expression class then), but query_class line should be here!

mvorisek: done, of course - the first one was refactored only, but GH is showing 1 red/removed line and 1 added

tests/ExceptionTest.php
```diff
@@ -57,9 +51,17 @@ public function testNonexistantFieldException()
...
+        $q->table('dual')->field('do_not_exist');
+
+        $this->setExpectedException(\atk4\dsql\ExecuteException::class);
 
-        $this->setExpectedException('atk4\dsql\ExecuteException');
-        $q->execute(); // PDOException: SQLSTATE[HY000]: General error: 1 no such table: foo
+        try {
+            $q->execute();
+        } catch (\atk4\dsql\ExecuteException $e) {
+            $this->assertSame(1054, $e->getCode());
```
mvorisek: This is my idea to test, but still do not know how to query MySQL/MariaDB here

DarkSide666: How about using `driver` property? It's `sqlite`, `mysql` or `pgsql` or `oracle` - taken from Persistence class.

DarkSide666 (approved):
LGTM

tests/ExceptionTest.php
mvorisek: Why not, but I can not really open hundreds of PRs for these smaller fixes.

PR #180, opened 2020-04-08
Revert "Improve DSQL exception for SQL"

DarkSide666:
Reverts atk4/dsql#174

Reverting because:
- need to remove unused `q()` method
- need to fix tests and see why it throws PDO exception instead of ExecutionException.


mvorisek:
Closing, fixed in #179

PR #182, opened 2020-04-09
Implemented class mapping

nitricware:
Hello,

I implemented `PDO::FETCH_CLASS` in a non-breaking, optional way

Compare with Example #4: https://www.php.net/manual/de/pdostatement.fetchall.php

This effects `get()`, `getOne()` and `getRow()` in `Expression.php`.

Maybe it's of use?

Thanks, Kurt

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -648,7 +648,7 @@ class Expression implements \ArrayAccess, \IteratorAggregate, ResultSet
      *
      * @return array
      */
-    public function get()
+    public function get(string $classname = 'stdClass')
     {
         $stmt = $this->execute();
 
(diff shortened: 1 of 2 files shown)


src/Expression.php
```diff
@@ -648,14 +648,18 @@ public function getIterator()
      *
      * @return array
      */
-    public function get()
+    public function get(string $classname = 'stdClass', $returnObject = false)
```
mvorisek, line 651: Seems like a bad design - why two parameters if one is enought? And you should use `\stdClass::class` instead of string.

mvorisek (changes requested):

mvorisek:
What is the real use case? Makes this even sense with dsql? (user can always use `execute()` + anything he needs)

nitricware:
Use case was comfort.

However, as you pointed out

```php
$e = $q->execute();
$r = $e->fetchAll(\PDO::FETCH_CLASS, SomeClass::class);
```

does the job just fine.

> why two parameters if one is enought? [sic!]

`$returnObject = false` is for compatibility reasons only. The tests would fail because at some point, an `array` is expected and an `object` is returned.

Since this pull request can be worked around with just one line of extra code, I'll close the pull request. Thanks for considering! :)

PR #183, opened 2020-04-09
Solve atk4/data first - Remove old Travis files

mvorisek:
If needed anytime, they can be restored from history, but having old/ununsed files in repo is misleading.


mvorisek:
Tests should be fixed once https://github.com/atk4/dsql/pull/179 is merged.

DarkSide666 (changes requested):
I'm OK to remove `.old.travis.yml` file, but these phpunit config files was used not only by travis, but also you can use them when running tests locally (on localhost). So I would like to keep them as they are.

Also tests still fail.

mvorisek:
> 
> 
> I'm OK to remove `.old.travis.yml` file, but these phpunit config files was used not only by travis, but also you can use them when running tests locally (on localhost). So I would like to keep them as they are.

So can we change the database hosts/configuratioon (should be possible) and use only one config file for local/GH?

I propose these changes as they are not self explanatory and it takes uneeded time get what is really does..

mvorisek:
Closing in favor of PR #190

PR #185, opened 2020-04-11
Refactor driver type

mvorisek:


DarkSide666:
What advantage it gives to just rename this property?
But overall I agree that `driverType` is more appropriate property name.

mvorisek:
> 
> 
> What advantage it gives to just rename this property?
> But overall I agree that `driverType` is more appropriate property name.

`driver` is misleading, it it not an object representing driver. Once this is merged, we can refactor in other repos.

DarkSide666 (approved):
Added more changes required. Now looks good.

mvorisek:
@DarkSide666 Perfect!

PR #195, opened 2020-04-29
Feature/pgsql exception

DarkSide666:


src/Connection_PgSQL.php
```diff
@@ -15,18 +17,22 @@ class Connection_PgSQL extends Connection
...
     public function lastInsertID($m = null)
     {
         // PostGRE SQL PDO requires sequence name in lastInertID method as parameter
         try {
-            return $this->connection()->lastInsertID($m->sequence ?: $m->table . '_' . $m->id_field . '_seq');
+            $seq = is_string($m) ? $m : ($m->sequence ?: $m->table . '_' . $m->id_field . '_seq');
+
+            return $this->connection()->lastInsertID($seq);
         } catch (\PDOException $e) {
+            throw $e;
```
mvorisek: Then simply do not catch the exceptions :)

src/Connection_PgSQL.php
```diff
@@ -15,18 +17,22 @@ class Connection_PgSQL extends Connection
      *
      * Few Connection drivers need to receive Model to get ID because PDO doesn't support this method.
      *
-     * @param \atk4\data\Model Optional data model from which to return last ID
+     * @param Model|string Optional data model from which to return last ID or sequence name
```
mvorisek: Types hell again, no! pass it as a 2nd param.

mvorisek (changes requested):

src/Connection.php
```diff
@@ -404,11 +406,12 @@ public function rollBack()
...
      * Few Connection drivers need to receive Model to get ID because PDO doesn't support this method.
      *
-     * @param \atk4\data\Model Optional data model from which to return last ID
+     * @param Model Optional data model from which to return last ID
+     * @param string Optional sequence name from which to return last ID (this takes precedence)
      *
      * @return mixed
      */
-    public function lastInsertID($m = null)
+    public function lastInsertID(Model $m = null, string $sequence = null)
```
mvorisek: We are at atk4/dsql here, 1st param should be completely removed.

mvorisek (changes requested):

src/Connection.php
DarkSide666: Hmm. In theory yes - atk4/dsql should not rely on atk4/data.
But on other hand - see Connection_Oracle and Connection_PgSQL. How can I get lastInsertID from $model->sequence if dsql will know nothing about models ?

mvorisek: https://github.com/atk4/dsql/pull/195/files#diff-ae503101b8b31c128532e45a73ea0dc0R28

this should be put in data/Model and passed here as a string, no other usage here

DarkSide666: OK, you're right. Will make a fix here and also small relates PR in data now.

src/Connection_Oracle.php
```diff
@@ -32,27 +30,19 @@ public function __construct($properties = [])
...
-            // Imants: Disabled for now because otherwise this will work even if database use triggers or
-            // any other mechanism to automatically increment ID and we can't tell this line to not execute.
-            //return $this->expr('SELECT max([field]) FROM [table]', ['field'=>$m->id_field, 'table'=>$m->table])->getOne();
+        if ($sequence) {
+            return $this->dsql()->mode('seq_currval')->sequence($sequence)->getOne();
         }
 
         // fallback
-        return parent::lastInsertID($m, $sequence);
+        return parent::lastInsertID($sequence);
```
mvorisek, line 46: No need for fallback, citing from PHP doc:
> PDO_PGSQL requires you to specify the name of a sequence object

mvorisek (changes requested):

src/Connection_Oracle.php
DarkSide666: This ir Oracle not PgSQL. So if we have sequence set, then we call SELECT $sequence.CURRVAL FROM dual otherwise fallback to default lastInsertID. For example, id is set by some DB trigger.

mvorisek: I do not know Oracle behaviour on this... If there is usecase, then ok, if not, it should be removed.

mvorisek (approved):

src/Connection_Oracle.php
DarkSide666: There can be. I have used Oracle some years ago (not with agile toolkit).

PR #196, opened 2020-04-29
TO BE SPLITTED: Refactor/organize folder structure

georgehristov:
- Move DB specific code to designated namespaces
- Introduce Connection class registry based on driverType
- Utilize object inheritance for setting up connection specific settings and behavior
- Improve variable naming consistency
- Change variable and method names based on semantics
- Keep full BC

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -54,17 +54,17 @@ class Connection
      *
      * @var array
      */
-    protected static $registry = [
-        'sqlite' => SQLite\Connection::class,
-        'mysql' => MySQL\Connection::class,
-        'pgsql' => PgSQL\Connection::class,
-        'oci' => Oracle\Connection::class,
+    protected static $driverRegistry = [
+        'sqlite' => SQLite\Driver::class,
+        'mysql' => MySQL\Driver::class,
+        'pgsql' => PgSQL\Driver::class,
+        'oci' => Oracle\Driver::class,
         // backward compatibility
-        'dumper' => Debug\Stopwatch\Connection::class,
-        'stopwatch' => Debug\Stopwatch\Connection::class,
+        'dumper' => Debug\Stopwatch\Driver::class,
+        'stopwatch' => Debug\Stopwatch\Driver::class,
         // backward compatibility
-        'counter' => Debug\Profiler\Connection::class,
-        'profile' => Debug\Profiler\Connection::class,
+        'counter' => Debug\Profiler\Driver::class,
+        'profile' => Debug\Profiler\Driver::class,
     ];
 
     /**
@@ -166,9 +166,9 @@ class Connection
             /**
              * @var Connection $connectionClass
              */
-            $connectionClass = self::resolveDriver($driverType);
+            $driverClass = self::resolveDriver($driverType);
 
-            return new $connectionClass(array_merge([
+            return new $driverClass(array_merge([
                 'driver' => $dsn,
             ], $args));
         }
@@ -186,10 +186,10 @@ class Connection
         /**
          * @var Connection $connectionClass
          */
-        $connectionClass = self::resolveDriver($dsn['driverType']);
+        $driverClass = self::resolveDriver($dsn['driverType']);
 
-        return new $connectionClass(array_merge([
-            'driver' => $connectionClass::createDriver($dsn),
+        return new $driverClass(array_merge([
+            'driver' => $driverClass::createDriver($dsn),
         ], $args));
     }
 
@@ -204,30 +204,30 @@ class Connection
      * CustomDriver\Connection must be descendant of Connection class.
      *
      * @param string $driverType
-     * @param string $connectionClass
+     * @param string $driverClass
      */
-    public static function register($driverType = null, $connectionClass = null)
+    public static function registerDriver($driverType = null, $driverClass = null)
     {
-        if (!$connectionClass && is_a($driverType, self::class, true)) {
-            $connectionClass = $driverType;
+        if (!$driverClass && is_a($driverType, self::class, true)) {
+            $driverClass = $driverType;
             $driverType = null;
         }
 
-        $connectionClass = $connectionClass ?? static::class;
+        $driverClass = $driverClass ?? static::class;
 
-        $driverType = $driverType ?? $connectionClass::defaultDriverType();
+        $driverType = $driverType ?? $driverClass::defaultDriverType();
 
         if (is_array($driverTypes = $driverType)) {
-            foreach ($driverTypes as $driverType => $connectionClass) {
+            foreach ($driverTypes as $driverType => $driverClass) {
                 if (is_numeric($driverType)) {
-                    $driverType = $connectionClass::defaultDriverType();
+                    $driverType = $driverClass::defaultDriverType();
                 }
 
-                static::register($driverType, $connectionClass);
+                static::registerDriver($driverType, $driverClass);
             }
         }
 
-        self::$registry[$driverType] = $connectionClass;
+        self::$driverRegistry[$driverType] = $driverClass;
     }
 
     /**
@@ -239,7 +239,7 @@ class Connection
      */
     public static function resolveDriver($driverType)
     {
-        return self::$registry[$driverType] ?? static::class;
+        return self::$driverRegistry[$driverType] ?? static::class;
     }
 
     /**
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -196,9 +196,9 @@ class Connection
             $connectionClass = $driverType;
             $driverType = null;
         }
-        
+
         $connectionClass = $connectionClass ?? static::class;
-        
+
         $driverType = $driverType ?? $connectionClass::DEFAULT_DRIVER_TYPE;
 
         if (is_array($driverTypes = $driverType)) {
@@ -206,7 +206,7 @@ class Connection
                 if (is_numeric($driverType)) {
                     $driverType = $connectionClass::DEFAULT_DRIVER_TYPE;
                 }
-                
+
                 static::register($driverType, $connectionClass);
             }
         }
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -42,12 +42,12 @@ class Connection
     public $driverType;
 
     protected static $registry = [
-            'sqlite' => SQLite\Connection::class,
-            'mysql'  => MySQL\Connection::class,
-            'pgsql'  => PgSQL\Connection::class,
-            'oci'    => Oracle\Connection::class,
-            'dumper' => Dumper\Connection::class,
-            'counter'=> Counter\Connection::class,
+        'sqlite' => SQLite\Connection::class,
+        'mysql' => MySQL\Connection::class,
+        'pgsql' => PgSQL\Connection::class,
+        'oci'    => Oracle\Connection::class,
+        'dumper' => Dumper\Connection::class,
+        'counter'=> Counter\Connection::class,
     ];
 
     /**
@@ -153,7 +153,7 @@ class Connection
             $connectionClass = self::resolve($driverType);
 
             return new $connectionClass(array_merge([
-                    'handler' => $dsn,
+                'handler' => $dsn,
             ], $args));
         }
 
@@ -173,12 +173,12 @@ class Connection
         $connectionClass = self::resolve($dsn['driverType']);
 
         return new $connectionClass(array_merge([
-                'handler' => $connectionClass::createHandler($dsn),
+            'handler' => $connectionClass::createHandler($dsn),
         ], $args));
     }
 
     /**
-     * Adds connection class to the registry for resolving in Connection::resolve method.
+     * Adds connection class to the registry for resolving in Connection::resolve method
      *
      * Can be used as:
      *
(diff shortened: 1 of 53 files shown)


src/Connection.php
```diff
@@ -9,17 +9,39 @@ class Connection
...
-    protected $expression_class = Expression::class;
+    protected $expressionClass = Expression::class;
 
     /** @var Connection|\PDO Connection or PDO object */
+    protected $handler;
+
+    /**
+     * @deprecated use $handler instead
+     */
     protected $connection;
```
mvorisek, line 50: This is protected, there does not need to be BC. We are not Google.

src/Connection.php
```diff
@@ -9,17 +9,39 @@ class Connection
...
 
     /** @var int Current depth of transaction */
-    public $transaction_depth = 0;
+    public $transactionDepth = 0;
+
+    /**
+     * @deprecated use $transactionDepth instead
+     */
+    public $transaction_depth;
```
mvorisek: If used somewhere, PHP will emit a notice that you try to use an unitialized property.

src/Connection.php
```diff
@@ -29,22 +51,38 @@ class Connection
...
     {
         if (!is_array($properties)) {
             throw new Exception([
-                'Invalid properties for "new Connection()". Did you mean to call Connection::connect()?',
+                'Invalid properties for "new Connection()". Did you mean to call Connection::create()?',
                 'properties' => $properties,
             ]);
         }
 
+        $this->driverType = static::DEFAULT_DRIVER_TYPE;
```
mvorisek: If default type should be null, then there should be not const like that.

src/Connection.php
```diff
@@ -97,7 +135,15 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
     }
 
     /**
-     * Connect database.
+     * @deprecated use Connection::create instead
+     */
+    public static function connect($dsn, $user = null, $password = null, $args = [])
```
mvorisek, line 143: With each meaningful deprecation I would you like to see when it should be removed. We need to reduce to the code as much as possible.

src/Connection.php
```diff
@@ -106,125 +152,98 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
      *
      * @return Connection
      */
-    public static function connect($dsn, $user = null, $password = null, $args = [])
+    public static function create($dsn, $user = null, $password = null, $args = [])
```
mvorisek, line 158: When you create a Pdo, does it try to connect immediatelly?

src/Connection.php
```diff
@@ -106,125 +152,98 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
...
+     * MySQL\Connection::register()
+     *
+     * CustomDriver\Connection must be descendant of Connection class.
+     *
+     * @param string $driverType
+     * @param string $connectionClass
+     */
+    public static function register($driverType = null, $connectionClass = null)
+    {
+        if (!$connectionClass && is_a($driverType, self::class, true)) {
```
mvorisek: This class can be abstract if needed. Do we need direct instance of Connection anywhere?

src/Connection.php
```diff
@@ -106,125 +152,98 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
...
+    }
+
+    /**
+     * Resolves the connection class to use based on driver type.
+     *
+     * @param string $driverType
+     *
+     * @return string
+     */
+    public static function resolve($driverType)
```
mvorisek: Too generic name

src/Connection.php
```diff
@@ -236,48 +255,56 @@ protected static function getPDO(array $dsn)
...
+        return $this->handler();
+    }
+
+    /**
+     * Returns the connection handler.
      *
      * @return Connection|\PDO
      */
-    public function connection()
+    public function handler()
```
mvorisek: Driver seems more industry common to me here.

Is it desireable to allow - use case? - `\Connection` here at all?

src/Counter/Connection.php
```diff
@@ -0,0 +1,113 @@
...
+
+namespace atk4\dsql\Counter;
+
+use atk4\dsql\Expression;
+use atk4\dsql\ProxyConnection;
+use atk4\dsql\Query;
+
+class Connection extends ProxyConnection
+{
+    const DEFAULT_DRIVER_TYPE = 'counter';
```
mvorisek: Why not simply DRIVER_TYPE ?

src/Expression.php
```diff
@@ -41,7 +41,7 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      *
      * @var string
      */
-    protected $escape_char = '"';
+    protected $quotedIdentifier = '"';
```
mvorisek: Identifier? `escapeChar` seems better to me, putting here something more than a char probably is wrong.

src/Connection.php
```diff
@@ -236,48 +255,56 @@ protected static function getPDO(array $dsn)
...
+        return $this->handler ?? $this;
     }
 
     /**
      * Execute Expression by using this connection.
      *
      * @return \PDOStatement
      */
-    public function execute(Expression $expr)
+    public function execute(Expression $expression)
```
mvorisek: On ther places there is `$expr`, usually I do not like shortening, but it makes sense here (keep the original).

mvorisek (changes requested):

src/Connection.php
georgehristov: The problem is `DIContainerTrait` lets setting protected properties from external locations.
Some of the legacy projects might use `['connection' => Whatever]` so I decided to keep it BC for time being.

src/Connection.php
georgehristov: `static::DEFAULT_DRIVER_TYPE` is defined in each child class

src/Connection.php
georgehristov: The semantics is `Connection::create`. So you create a connection and it connects

src/Expression.php
georgehristov: This is not an escape string but quoted identifier.
https://dba.stackexchange.com/questions/23129/benefits-of-using-backtick-in-mysql-queries

src/Counter/Connection.php
georgehristov: The actual driver type is in `$driverType` property.
`DEFAULT_DRIVER_TYPE` is the default one that will be used if no other driverType is set
I spend a few hours figuring out how to name it. Naming it `DRIVER_TYPE` does not hint to what the actual use is and what is the difference with the property `$driverType`

src/Connection.php
mvorisek: Wdyt to update `DIContainerTrait` which throw an exception (with hint) when deprecated property is set with it?

Or even a new core trait which will accept list of deprecated properties/methods and throw an exception when any of them are accessed (using magic methods/already existing trait).

src/Connection.php
mvorisek: I am okwith it, only a question if the rename is worth enought.

src/Counter/Connection.php
mvorisek: What is the problem with setting `$driverType` to the default value?

src/Connection.php
georgehristov: Using `handler` here also needed some thought. Handler can be a `Connection` and `PDO` or some other class so it has use cases and does not confuse the dev having `Connection::connection`
Handler handles the connection

src/Expression.php
mvorisek: Is this guarantted to be never used with values and only with names?

src/Connection.php
mvorisek: I got confused myself in the past... Ok to rename to `driver`?

src/Connection.php
georgehristov: Yes, good idea!

src/Counter/Connection.php
georgehristov: My idea was to have a way of retrieving the default driver type for a connection class without having to instantiate it. (e.g `MySQL\Connection::DEFAULT_DRIVER_TYPE`)
This is used in `Connection::register` where the connection class is registered for its default driver type if no other set specifically.

mvorisek: I am not sure if that can be done using reflection, but probably yes :D

mvorisek: https://www.php.net/manual/en/reflectionclass.getdefaultproperties.php

src/Connection.php
georgehristov: Reverted

src/Counter/Connection.php
georgehristov: Used your suggestion and got rid of the constant altogether

src/Connection.php
georgehristov: Renamed to driver now if it is more understandable

src/Connection.php
georgehristov: Renamed to `resolveDriver`

src/Expression.php
georgehristov: This string is used to wrap (quote) literal strings to be used in the query like in MySQL
```
CREATE TABLE `my table` (id INT);
```

mvorisek: If, then `quoteIdentifier`.

But Microsoft https://docs.microsoft.com/en-us/sql/t-sql/statements/set-quoted-identifier-transact-sql uses this term for string values not string identifiers..

src/Counter/Connection.php
mvorisek: 👍

src/Connection.php
mvorisek: 👍 wdyt about the same for registry, register() method..?

georgehristov: I believe keeping only `register` is more clear and understandable.
`CustomConnection::register();` or
`CustomConnection::register('forSomeDriverType');` or 
`Connection::register('someDriverType', CustomConnection::class)`

mvorisek: I meant `registerDriver`, dyt it is less clear?

georgehristov: Yes, I understood about `registerDriver`. I think now is more clear but let me think a bit more about it.

mvorisek: Or even `registerDriverClass`.

georgehristov: Now is reformatted and more understandable I believe

src/Connection_Oracle.php
```diff
@@ -3,8 +3,8 @@
 namespace atk4\dsql;
 
 /**
- * @deprecated use Oracle\Connection instead
+ * @deprecated use Oracle\Driver instead
  */
-class Connection_Oracle extends Oracle\Connection
+class Connection_Oracle extends Oracle\Driver
```
mvorisek, line 8: Is it better? I meant the to only refactor the handler to driver, driver term is usually used for `\Pdo` driver.

georgehristov: The thing is that an object of Connection class can also be a driver.
So the DB specific implementation of Connection we call Driver and we register it

mvorisek: I see... But it extend `Connection`, so using 2 terms for the same seems not good to me :)

georgehristov: The mice are playing but let's see what the big cats @romaninsh and @DarkSide666 will say about all this :)

src/MySQL/Driver.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+namespace atk4\dsql\MySQL;
+
+use atk4\dsql\Connection;
+
+class Driver extends Connection
+{
+    public $driverType = 'mysql';
```
mvorisek, line 9: What is the actual purpose of driverType, when we can ceck if the class is instanceof? Should be probably removed.

georgehristov: Refactored a bit more https://github.com/atk4/dsql/tree/refactor/organize-folder-structure-driver

WDYT?

DarkSide666:
Huge amount of backward incompatible things :(

georgehristov:
> 
> 
> Huge amount of backward incompatible things :(

This is not the idea. In my opinion should be BC but needs to be tested on actual projects

mvorisek:
@georgehristov can you update this PR to start at current develop without any folder renames? (we can check and decise with folder structure rearrange later)

georgehristov:
Folder structure is major part of the refactoring. And all is intended / tested to be BC.
I have similar folder restructure lined up for the other repos as well.

mvorisek:
@georgehristov There were many changes, this is primary to do a review, see real changes and then we can reupdate the structure very easily.

mvorisek:
@georgehristov PRs like this are bad. They can not be checked nor merge if not done so immediatelly.

Is it ok with you to split this PR at least into two - one for pure folder/NS refactor and another for changes?

src/MySQL/Driver.php
mvorisek: @georgehristov 

1. huge 👍 for splitting this PR. In the future, it is really better to isolate larger changes into more isolated oned, good job!
2. is dsql/218 the last part?
3. here are some comments to which seems to be not addressed

at least:
- about the `driverType` - is seems fully redundant for and I think we should test `instanceof Engine/Connection` to the right type (also much more refactorable/usable)
- in the new PR you use still `driver` word
- is it guaranteed that `create` does only create the underlaying driver and does not `connect`? (as with connect - I count with DB ping/issues, and with create - I expect no connection error/exception)

PR #198, opened 2020-04-30
make php-cs-fixer suitable for Windows as well

georgehristov:
Can be distributed to other repos as well

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -52,6 +52,8 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
+        'line_ending' => true,
     ])
     ->setFinder($finder)
-    ->setCacheFile(__DIR__ . '/.php_cs.cache');
+    ->setCacheFile(__DIR__ . '/.php_cs.cache')
+    ->setLineEnding("\r\n");


.php_cs.dist
```diff
@@ -52,6 +52,8 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
+        'line_ending' => true,
     ])
     ->setFinder($finder)
-    ->setCacheFile(__DIR__ . '/.php_cs.cache');
+    ->setCacheFile(__DIR__ . '/.php_cs.cache')
+    ->setLineEnding("\r\n");
```
mvorisek, line 59: Never ever CRLF!!! Reconfigure your IDE/git to use LF.

mvorisek (changes requested):

.php_cs.dist
```diff
@@ -52,6 +52,8 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
+        'line_ending' => true,
```
mvorisek, line 55: This is implied by the `@PSR2`, `@Symfony`, `@PhpCsFixer` rulesets, do not mess these configs with unneeded rules. I spend a lot of time to provide clean, strict, but still not too strict rules.

CRLF is also direct violation of PSR-12 standard https://www.php-fig.org/psr/psr-12/#22-files . The same probably for other languages :)

PR #201, opened 2020-05-01
Add escaping support to parser

### LINKED ISSUE #144: Escaping in expressions
Maybe we should have some way how to escape `{}` and `[]` parenthesis in expression strings? For example, in standard way - with backslash `\{...\}` and `\[...\]`.

Related question/answer/work-around here: https://stackoverflow.com/a/50759511/1466341

mvorisek:
fix https://github.com/atk4/dsql/issues/144

BC break: backslash escapes next character


mvorisek:
add tests, otherwise fully done

if expressions are somewhere built, generator need to be updates

PR #202, opened 2020-05-01
Run tests with bool column

mvorisek:


DarkSide666:
Please no `bool` here because PgSQL will need to cast values to bool then and will fail tests. Casting for PgSQL is not implemented because treatment of datatypes in PgSQL is much more strict than in MySQL.

mvorisek:
@DarkSide666 Experiment, maybe there is easy way to solve this. In this stage of course draft and DO NOT MERGE!

mvorisek:
@DarkSide666 can you merge this https://github.com/atk4/data/pull/571 as promised "after a hour" days ago?

DarkSide666:
yeah, sorry. had to work on projects too to earn some cash for living :)

mvorisek:
@DarkSide666 make Atk big and you are safe! :D

DarkSide666:
I doubt that toolkit as such (as framework) can be profitable.
Projects built on top of toolkit can.

DarkSide666 (approved):
Good way to fix this :)

PR #203, opened 2020-05-02
Deduplicate value cast and add support for bool

mvorisek:
For robustness and future, no bool type usage so far.


DarkSide666 (approved):
no more code duplication - good.

PR #204, opened 2020-05-15
Use MB functions

DarkSide666:


src/Connection.php
```diff
@@ -82,14 +82,14 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
 
         // If it's string, then find driver
         if (is_string($dsn)) {
-            if (strpos($dsn, ':') === false) {
+            if (mb_strpos($dsn, ':') === false) {
```
mvorisek: absolutely no need

src/Connection.php
```diff
@@ -70,7 +70,7 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
             // DSN is using URL-like format, so we need to convert it
             $dsn = $parts['scheme'] . ':host=' . $parts['host']
                 . (isset($parts['port']) ? ';port=' . $parts['port'] : '')
-                . ';dbname=' . substr($parts['path'], 1);
+                . ';dbname=' . mb_substr($parts['path'], 1);
```
mvorisek: Probably to exclude `/` as the first char, so no need.

src/Expression.php
```diff
@@ -587,7 +587,7 @@ public function execute(object $connection = null)
                     // workaround to support LOB data type 2/2, see https://github.com/doctrine/dbal/pull/2434
                     $statement->boundValues[$key] = $val;
                     if ($type === \PDO::PARAM_STR) {
-                        $bind = $statement->bindParam($key, $statement->boundValues[$key], $type, strlen((string) $val));
+                        $bind = $statement->bindParam($key, $statement->boundValues[$key], $type, mb_strlen((string) $val));
```
mvorisek: Be very carefull here, it may require a binary length here!!

src/Expression.php
```diff
@@ -424,14 +424,14 @@ public function render()
             //     param     |  escape   |  escapeSoft
             '/\[[a-z0-9_]*\]|{[a-z0-9_]*}|{{[a-z0-9_]*}}/i',
             function ($matches) use (&$nameless_count) {
-                $identifier = substr($matches[0], 1, -1);
+                $identifier = mb_substr($matches[0], 1, -1);
```
mvorisek: Example of misleading usage - this strips brackets from the sides...

mvorisek (changes requested):
Running CS fixer is fine, but you usually find 1, 2 valid usecases with these fixers... That is also why I reviewed the fixed files before applaying the most risky rules when added the global rules...

src/Expression.php
DarkSide666: yes most likely it should be binary

src/Expression.php
```diff
@@ -431,7 +431,7 @@ function ($matches) use (&$nameless_count) {
                 } elseif ($matches[0][0] === '{') {
                     if ($matches[0][1] === '{') {
                         $escaping = 'soft-escape';
-                        $identifier = substr($identifier, 1, -1);
+                        $identifier = mb_substr($identifier, 1, -1);
```
mvorisek: misleading only...

mvorisek (changes requested):

src/Expression.php
DarkSide666: ok

DarkSide666:
OK, nothing is actually left there. Only driverType which also should never be multibyte.
So I guess this PR can be closed without merging :)

src/Expression.php
mvorisek: I do not know, should be checked, it is probably for maximal byte length to allocate, my guess.

PR #213, opened 2020-06-19
[refactor] introduce debug classes under new structure

georgehristov:
Split of https://github.com/atk4/dsql/pull/196

Part I - Refactoring of Debug proxy connections

### DIFF
--- a/src/Connection_Counter.php
+++ b/src/Connection_Counter.php
@@ -5,8 +5,8 @@ declare(strict_types=1);
 namespace atk4\dsql;
 
 /**
- * @deprecated will be removed in v2.3
+ * @deprecated will be removed in dec-2020
  */
-class Connection_Counter extends Debug\Profiler\Driver
+class Connection_Counter extends Debug\Profiler\Connection
 {
 }
--- a/src/Connection_Counter.php
+++ b/src/Connection_Counter.php
@@ -4,101 +4,10 @@ declare(strict_types=1);
 
 namespace atk4\dsql;
 
-class Connection_Counter extends Connection_Proxy
+/**
+ * @deprecated will be removed in v2.3
+ */
+class Connection_Counter extends Debug\Profiler\Driver
 {
-    /**
-     * Callable to call for outputting.
-     *
-     * Will receive parameters:
-     *  - int     Count of executed queries
-     *  - int     Count of executed selects
-     *  - int     Count of rows iterated
-     *  - int     Count of executed expressions
-     *  - boolean True if we had exception while executing expression
-     *
-     * @var callable
-     */
-    public $callback;
-
-    /** @var int Count of executed selects */
-    protected $selects = 0;
-
-    /** @var int Count of executed queries */
-    protected $queries = 0;
-
-    /** @var int Count of executed expressions */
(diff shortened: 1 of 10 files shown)


src/Connection_Counter.php
```diff
@@ -4,101 +4,9 @@
 
 namespace atk4\dsql;
 
-class Connection_Counter extends Connection_Proxy
+/**
+ * @deprecated will be removed in v2.3
```
mvorisek: I propose to write here: "will be removed in 2020-dec"

It is much easier for users to know when it will be removed and when to remove it in the code.

(update in everywhere)

mvorisek:
Seems ok to merge except one thing - Driver vs. Connection.

Why is driver really that much better?

Driver is usually used for `mysqli`, `PDO`, so `Connection` seems here ok for me to keep and to keep all consistency.

Even if really needed, I propose to keep `Connection` for now and rename everything at once after the new structure is set. Why? It is much easier to refactor purely NS in user's projects and later (after another consistent release) refactor the short name.

@georgehristov ok with you?

src/Debug/Stopwatch/Connection.php
```diff
@@ -0,0 +1,70 @@
...
+     * @var callable
+     */
+    public $callback;
+
+    /**
+     * @var float
+     */
+    protected $startTime;
+
+    public static function createDriver(array $dsn)
```
mvorisek: there is 2x `createDriver` - should be probably part of later PR when all other Connection NSes are updated and if we choose to use the name `Driver`.

georgehristov: Agree

mvorisek:
Then feel free to rename NS of other engines. But please, do not copy from old PR, only minimalistic rename, I will check (still have to compare manually because of old classes are not removed completely so git does handle this as renames).

georgehristov:
I would keep this PR only for the Debug connections.
You are right that the more concentrated is the PR the easier to understand and accept.
Refactoring of NS structure will follow in another PR.

src/Debug/Stopwatch/Connection.php
```diff
@@ -0,0 +1,65 @@
...
+
+            throw $e;
+        }
+
+        return $ret;
+    }
+
+    protected function dump(Expression $expr, $error = false)
+    {
+        $error = $error ? 'ERROR' : '';
```
mvorisek: To be consistent with the other class `'ERROR '` (with space at the end) and no space below. Otherwise PR for Debug classes refactor complete.

georgehristov:
@mvorisek done so pls approve for the record. I will not merge

mvorisek (approved):
Pure NS updates/renames are done well. Approving rename of Debug Connections. No unresolved feedback left.

Please rename freshly (do not reuse anything from changed code from your huge PR) other engines.

mvorisek:
@georgehristov I checked the changelog after rel 2.1.0 and we can merge this right away. If ok with you, merge this PR and create another one for the pure renames (with BC classes like here). @romaninsh is slow, we can merge all renames today today.

georgehristov:
Good. Moving on to next task now ...

PR #215, opened 2020-06-19
Rename classes - use NS instead of flat structure with "_"

georgehristov:
Split of https://github.com/atk4/dsql/pull/196

Part II - Pure refactoring of namespaces

### BC is provided for a short time, but not fully BC because `instanceof` operator like: `instanceof Query_MySQL` will fail with object instanced using the new naming

### DIFF
--- a/src/Connection_Counter.php
+++ b/src/Connection_Counter.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 namespace atk4\dsql;
 
+@trigger_error('Use atk4\dsql\Debug\Profiler\Connection instead', E_USER_DEPRECATED);
+
 /**
  * @deprecated will be removed in dec-2020
  */
--- a/src/Connection_Counter.php
+++ b/src/Connection_Counter.php
@@ -4,7 +4,9 @@ declare(strict_types=1);
 
 namespace atk4\dsql;
 
-@trigger_error('Use atk4\dsql\Debug\Profiler\Connection instead', E_USER_DEPRECATED);
+if (!class_exists(\SebastianBergmann\CodeCoverage\CodeCoverage::class, false)) {
+    'trigger_error'('Use atk4\dsql\Debug\Profiler\Connection instead', E_USER_DEPRECATED);
+}
 
 /**
  * @deprecated will be removed in dec-2020
--- a/tests/OracleTest.php
+++ b/tests/OracleTest.php
@@ -34,11 +34,11 @@ class OracleTest extends AtkPhpunit\TestCase
 
     public function connect($ver = '')
     {
-        $version = $ver ? "\\Version{$ver}" : '';
+        $version = $ver ? '\Version' . $ver : '';
 
         return new \atk4\dsql\Connection(array_merge([
             'connection' => new \PDO('sqlite::memory:'),
-            'query_class' => "\\atk4\\dsql\\Oracle{$version}\\Query",
+            'query_class' => '\atk4\dsql\Oracle' . $version . 'Query',
         ]));
     }
 
--- a/tests/OracleTest.php
+++ b/tests/OracleTest.php
@@ -36,7 +36,7 @@ class OracleTest extends AtkPhpunit\TestCase
     {
         return new \atk4\dsql\Connection(array_merge([
             'connection' => new \PDO('sqlite::memory:'),
-            'query_class' => \atk4\dsql\Query_Oracle::class . $ver,
+            'query_class' => \atk4\dsql\Oracle\Query::class . $ver,
         ]));
     }
 
--- a/tests/OracleTest.php
+++ b/tests/OracleTest.php
@@ -34,9 +34,11 @@ class OracleTest extends AtkPhpunit\TestCase
 
     public function connect($ver = '')
     {
+        $version = $ver ? "\\Version{$ver}" : '';
+
         return new \atk4\dsql\Connection(array_merge([
             'connection' => new \PDO('sqlite::memory:'),
-            'query_class' => \atk4\dsql\Oracle\Query::class . $ver,
+            'query_class' => "\\atk4\\dsql\\Oracle{$version}\\Query",
         ]));
     }
 
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -7,11 +7,11 @@ namespace atk4\dsql\tests;
 use atk4\core\AtkPhpunit;
 use atk4\dsql\Exception;
 use atk4\dsql\Expression;
-use atk4\dsql\MySQL;
+use atk4\dsql\Mysql;
 use atk4\dsql\Oracle;
-use atk4\dsql\PgSQL;
+use atk4\dsql\Postgresql;
 use atk4\dsql\Query;
-use atk4\dsql\SQLite;
+use atk4\dsql\Sqlite;
 
 /**
  * @coversDefaultClass \atk4\dsql\Query
@@ -1189,7 +1189,7 @@ class QueryTest extends AtkPhpunit\TestCase
      */
     public function testGroupConcat()
     {
-        $q = new MySQL\Query();
+        $q = new Mysql\Query();
         $this->assertSame('group_concat(`foo` separator :a)', $q->groupConcat('foo', '-')->render());
 
         $q = new Oracle\Query();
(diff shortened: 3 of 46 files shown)


tests/OracleTest.php
```diff
@@ -34,9 +34,11 @@ public function testDetection()
 
     public function connect($ver = '')
     {
+        $version = $ver ? "\\Version{$ver}" : '';
+
         return new \atk4\dsql\Connection(array_merge([
             'connection' => new \PDO('sqlite::memory:'),
-            'query_class' => \atk4\dsql\Query_Oracle::class . $ver,
+            'query_class' => "\\atk4\\dsql\\Oracle{$version}\\Query",
```
mvorisek: you use single quotes string here...

for prod/non test I qould use probably something like `['' => ::class, ...][$ver]`...

georgehristov: What do you mean single quotes string?

mvorisek:
@georgehristov rebase

image

why is there `c` suffix after the version?

tests/OracleTest.php
mvorisek: `'\atk4\dsql\Oracle' . $version . '...` should be the preffered format

- we should use single quoted strings everywhere
- do not escape `\` when not needed

georgehristov:
@mvorisek that is how it was referred to in the original class

mvorisek:
@georgehristov seems like typo... Connection_Oracle12 is/was without it!

georgehristov:
@mvorisek changed now

georgehristov:
https://www.oracle.com/database/12c-database/

@mvorisek so it should be `Version12c` everywhere

tests/QueryTest.php
```diff
@@ -7,12 +7,12 @@
 use atk4\core\AtkPhpunit;
 use atk4\dsql\Exception;
 use atk4\dsql\Expression;
+use atk4\dsql\MySQL\Query as Query_MySQL;
```
mvorisek: `use atk4\dsql\Mysql` shoud be preffered instead. `as Query_MySQL` is too chatty without any value provided and hard to refactor

also fix RandomTest (and maybe other) to use ...\tests NS.

mvorisek:
@georgehristov 
image

from Wiki, I would keep it without the suffix, just the version :)

Will be also consitent if we will need more DB specific classes for other engines.

georgehristov:
@mvorisek changed now, officially they refer to it with the "c" for whatever reason so we follow official version name

mvorisek:
No - see my update above, we want minimalistic name and consistency across classes

tests/QueryTest.php
georgehristov: What do you mean with `use atk4\dsql\Mysql`?

mvorisek: later refer to this import like `MySql\Connection` :) (never use aliases with _ if you can import the parent NS)

georgehristov: Beautiful 😄

mvorisek:
becase of https://github.com/atk4/ui/issues/1024 also rename the names to PascalCase:
- `MySQL` -> `Mysql`
- `SQLite` => `Sqlite`
etc. (the rule is - use letter capital when you will use it in snake case - ex. `mysql`, not `my_s_q_l`)

georgehristov:
This we discussed already and we keep like it is for now

mvorisek:
where - because it is planned like I propose :D

It would be better to do it now at once...

georgehristov:
> where - because it is planned like I propose :D

where?

mvorisek:
> we discussed already

georgehristov:
I will not change it now as I do not like the Pascal case in this instance (DB names)
If agreed by all then we can make another PR.

georgehristov:
> where - because it is planned like I propose :D

Where: Discord
When: 04/28/2020

mvorisek:
image

I see approval to go ahead with "snake case reasoning" :)

so is this ok with you to do now so we can merge today?

mvorisek:
- so ok with you to do it now?
- I also updated the desc - this PR is breaking and it may be very hard to discover it - ok to emit deprecated messages from the old classes when they are loaded?

georgehristov:
Actually I don't like it but I would follow if there is enough support for it.
But having @Darkside666 say it is maybe okay is not sufficient to set the direction.
So this can be done at later stage or we discuss here before merging.

mvorisek:
it is discussed - we should follow snake case reasoning, because not following it results in naming mess like `MySQLURL` instead of `MysqlUrl` (or mysql_url in snakecase, that is how would you normally also written it)

so this PR is complete except:
- [ ] naming normalization with snake_case rationale
- [x] emit deprecation warning in every deprecated class

src/Concerns/IsDeprecated.php
```diff
@@ -0,0 +1,18 @@
...
+namespace atk4\dsql\Concerns;
+
+/**
+ * use to trigger deprecated warning on a class.
+ */
+trait IsDeprecated
+{
+    public function __construct()
+    {
+        @trigger_error('Use ' . parent::class . ' instead', E_USER_DEPRECATED);
```
mvorisek: why @ before, will it actually emit anything?

If constructor is not used, this will not trigger... I think it is better to move it before/after the class def directly and emit these warnings always when the class is inluded.

georgehristov: The @ is added by CS Fixer.

I will move it in the file

src/Connection_Counter.php
```diff
@@ -4,6 +4,8 @@
 
 namespace atk4\dsql;
 
+@trigger_error('Use atk4\dsql\Debug\Profiler\Connection instead', E_USER_DEPRECATED);
```
mvorisek: You should always review CS fixes manually! I tested it now and with @ is does not emit any message!

If needed, write it as `'trigger_error'(...` (yes, with single quotes)

georgehristov: It does not work without @ either as warning is triggered immediately during testing caused by phpunit\php-code-coverage

mvorisek: https://3v4l.org/QepiX

georgehristov: This is clear. The problem is that the files are autoloaded by phpunit coverage so the warning is triggered during testing always

mvorisek: Does my "test" commit solves that?

mvorisek:
> naming normalization with snake_case rationale

@georgehristov can you finish this based on the discussion above? Exlc. this I am for merging this immediatelly.

src/Postgre/Connection.php
```diff
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace atk4\dsql\PgSQL;
+namespace atk4\dsql\Postgre;
```
mvorisek: I just written my friend working with this database daily. He told me that only two names are used for this DB - Postgres and Postgresql. I also asked him about PgSQL -> Pgsql, but he told me, that they almost never use this name for it.

Postgres is an old/alias name for Postgresql based on Wiki. So I would you full "Postgresql" here. Wdyt?

georgehristov: Aim is to be as intuitive and standard as possible and build the magic on top of that so `Postgresql` is fine with me.

mvorisek (approved):
Rename NS `Postgre` to `Postgresql` and merge. Good move for Atk, good work, @georgehristov!

(and update all dependent projects to use the new names)

src/Postgre/Connection.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+declare(strict_types=1);
+
+namespace atk4\dsql\Postgresql;
```
mvorisek: filepath as well...

georgehristov:
@mvorisek thanks for review and approval. I will not hurry to merge but await review/approval from @DarkSide666 as well

DarkSide666:
Sorry guys, it's Midsummer holidays here :)
I will quickly review now.

About naming Mysql vs MySQL - as I said, I like MySQL better, but I see the point of @mvorisek. Probably for the future when hopefully we will support like 30-50 different DB engines it's better to set quite strict (even if not intuitive at first sight) naming principles.

DarkSide666 (approved):
OK I think it's fine.

georgehristov:
@DarkSide666 at the moment in your code you can still use `MySQL\Connection` for instance (as I also prefer it being visually pleasing) as case is ignored.

mvorisek:
No you can not! Because of CS autoloader on non Windows platforms!

PR #216, opened 2020-06-19
Expression::getDebugQuery() no longer accepts $html param

georgehristov:


src/Expression.php
```diff
@@ -468,42 +468,27 @@ function ($matches) use (&$nameless_count) {
      */
     public function getDebugQuery($html = null)
```
mvorisek: better no HTML here, `SqlFormatter` is sudgested anyway.

and should be fixed in data/ui

georgehristov: You mean to deprecate the `$html` argument?

mvorisek: yes, completely remove - in this PR is not for NS rename ;-)

as it might imply a security risk if used somewhere, you can add there a check like

```
if (func_get_args_count() > 0) { // remove in 2020-dec
throw ..;
}
```

mvorisek (approved):
@georgehristov I have to say this PR was very poorly made and lot of fixes were needed.

PR #218, opened 2020-06-22
Add Connection classes registry (instead of hardcoded list)

georgehristov:
full BC

Avoid hardcoding of driver types and introduces possibility to extend the list without modifying the class

### DIFF
--- a/docs/connection.rst
+++ b/docs/connection.rst
@@ -16,10 +16,10 @@ Connection class is handy to have if you plan on building and executing
 queries in your application. It's more appropriate to store
 connection in a global variable or global class::
 
-    $app->db = atk4\dsql\Connection::establish($dsn, $user, $pass);
+    $app->db = atk4\dsql\Connection::connect($dsn, $user, $pass);
 
 
-.. php:staticmethod:: establish($dsn, $user = null, $password = null, $args = [])
+.. php:staticmethod:: connect($dsn, $user = null, $password = null, $args = [])
 
     Determine which Connection class should be used for specified $dsn,
     establish connection to DB by creating new object of this connection class and return.
@@ -61,11 +61,11 @@ Here is how you can use all of this together::
 
     $dsn = 'mysql:host=localhost;port=3307;dbname=testdb';
 
-    $connection = atk4\dsql\Connection::establish($dsn, 'root', 'root');
+    $connection = atk4\dsql\Connection::connect($dsn, 'root', 'root');
 
     echo "Time now is : ". $connection->expr("select now()");
 
-:php:meth:`establish` will determine appropriate class that can be used for this
+:php:meth:`connect` will determine appropriate class that can be used for this
 DSN string. This can be a PDO class or it may try to use a 3rd party connection
 class.
 
@@ -79,22 +79,22 @@ if you connect to vendor that does not use PDO.
     :param Expression  $expr: Expression (or query) to execute
     :returns: PDOStatement, Iterable object or Generator.
     
-.. php:method:: registerDriver($driverType = null, $connectionClass = null)
+.. php:method:: registerConnection($connectionType = null, $connectionClass = null)
 
-    Adds connection class to the registry for resolving in Connection::resolve method.
+    Adds connection class to the registry for resolving in Connection::resolveConnection method.
 
-    :param string $driverType Alias of the driver
+    :param string $connectionType Alias of the connection
     :param string $connectionClass The connection class to be used for the diver type
 
-Developers can register custom classes to handle driver types using the `Connecion::registerDriver` method::
+Developers can register custom classes to handle driver types using the `Connecion::registerConnection` method::
 
-   Connection::register('mysql', Custom\MySQL\Connection::class); // or directly using the class
-   Custom\MySQL\Connection::register();
+   Connection::registerConnection('mysql', Custom\MySQL\Connection::class); // or directly using the class
+   Custom\MySQL\Connection::registerConnection();
    
 The driver type used in the latter case is the default value of the `$driverType` property of 
 `Custom\MySQL\Connection`
 
-.. php:method:: establishDriverConnection(array $dsn)
+.. php:method:: establishConnection(array $dsn)
 
    The method should establish connection with DB and return the underlying connection object used by 
    the `Connection` class. By default PDO is used but the method can be overriden to return custom object to be 
--- a/docs/connection.rst
+++ b/docs/connection.rst
@@ -59,7 +59,6 @@ new Query or Expression class::
 
 Here is how you can use all of this together::
 
-
     $dsn = 'mysql:host=localhost;port=3307;dbname=testdb';
 
     $connection = atk4\dsql\Connection::establish($dsn, 'root', 'root');
@@ -79,3 +78,22 @@ if you connect to vendor that does not use PDO.
 
     :param Expression  $expr: Expression (or query) to execute
     :returns: PDOStatement, Iterable object or Generator.
+    
+.. php:method:: registerDriver($driverType = null, $connectionClass = null)
+
+    Adds connection class to the registry for resolving in Connection::resolve method.
+
+    :param string $driverType Alias of the driver
+    :param string $connectionClass The connection class to be used for the diver type
+
+Developers can register custom classes to handle driver types using the `Connecion::registerDriver` method::
+
+   Connection::register('mysql', Custom\MySQL\Connection::class); // or directly using the class
+   Custom\MySQL\Connection::register();
+   
+The driver type used in the latter case is the default value of the `$driverType` property of `Custom\MySQL\Connection`
+
+.. php:method:: createDriverConnection(array $dsn)
+
+   The method should return the underlying connection object used by the `Connection` class. By default PDO is used but
+   the method can be overriden to return custom object to be used for connection to DB.
--- a/docs/connection.rst
+++ b/docs/connection.rst
@@ -19,10 +19,10 @@ connection in a global variable or global class::
     $app->db = atk4\dsql\Connection::establish($dsn, $user, $pass);
 
 
-.. php:staticmethod:: connect($dsn, $user = null, $password = null, $args = [])
+.. php:staticmethod:: establish($dsn, $user = null, $password = null, $args = [])
 
     Determine which Connection class should be used for specified $dsn,
-    create new object of this connection class and return.
+    establish connection to DB by creating new object of this connection class and return.
 
     :param string $dsn: DSN, see http://php.net/manual/en/ref.pdo-mysql.connection.php
     :param string $user: username
@@ -91,9 +91,11 @@ Developers can register custom classes to handle driver types using the `Conneci
    Connection::register('mysql', Custom\MySQL\Connection::class); // or directly using the class
    Custom\MySQL\Connection::register();
    
-The driver type used in the latter case is the default value of the `$driverType` property of `Custom\MySQL\Connection`
+The driver type used in the latter case is the default value of the `$driverType` property of 
+`Custom\MySQL\Connection`
 
-.. php:method:: createDriverConnection(array $dsn)
+.. php:method:: establishDriverConnection(array $dsn)
 
-   The method should return the underlying connection object used by the `Connection` class. By default PDO is used but
-   the method can be overriden to return custom object to be used for connection to DB.
+   The method should establish connection with DB and return the underlying connection object used by 
+   the `Connection` class. By default PDO is used but the method can be overriden to return custom object to be 
+   used for connection to DB.
--- a/docs/connection.rst
+++ b/docs/connection.rst
@@ -79,17 +79,17 @@ if you connect to vendor that does not use PDO.
     :param Expression  $expr: Expression (or query) to execute
     :returns: PDOStatement, Iterable object or Generator.
     
-.. php:method:: registerConnection($connectionClass = null, $connectionType = null)
+.. php:method:: registerConnectionClass($connectionClass = null, $connectionType = null)
 
-    Adds connection class to the registry for resolving in Connection::resolveConnection method.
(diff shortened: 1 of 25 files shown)


src/Connection.php
```diff
@@ -31,6 +31,22 @@ class Connection
      */
     public $driverType;
 
+    /**
```
mvorisek, line 34: commented here: https://github.com/atk4/dsql/pull/196#discussion_r443658293

`driverType` can make sense if connection is wrapped, but I think we does not even support this scenario now (and we should not, we should support transparent layering)

for this reaon, I think we can deprecate this property in advance to native `instanceof` operator

romaninsh:
Would be nice if it had more examples how to register new drivers...

georgehristov:
I will add in docs...

georgehristov:
@mvorisek moving discussion here

> - about the driverType - is seems fully redundant for and I think we should test instanceof Engine/Connection to the right type (also much more refactorable/usable)

I will think about BC solution but I prefer using a method like `$connection->usesDriver($someClassOrAlias)` instead of pure `instanceof` when necessary

> - in the new PR you use still driver word

Where?
If you are referring to methods/variables the challenge with dsql is that `Connection` is recursive and can contain other connection in the `$connection` property.

> - is it guaranteed that create does only create the underlying driver and does not connect? (as with connect - I count with DB ping/issues, and with create - I expect no connection error/exception)

Create connects as well (same as the deprecated `Connection::connect`method)
I simply do not find it intuitive to connect a connection. You create a connection (it connects and returns the connection object). The other word that comes to mind is "establish a connection"
Ideas and comments are welcome.

mvorisek:
@georgehristov 

`usesDriver` why? because you do not have to extend from single engine class? In theory yes, but Atk is not much (actually zero) abstract/interface based... So for now, I would go with the simplest solution - drop the driverType solution, it is used more or less in tests only.

Where? driverRegistry, resolveDriver... (in Atk it should be more seedByScheme?)

I get it. You do not want `connect` because it "also create". So maybe `createAndConnect`? I am really for keeping the word connect there to make it clear what it does. Reseach might help, do not reinvent the wheel.

docs/connection.rst
```diff
@@ -91,9 +91,11 @@ Developers can register custom classes to handle driver types using the `Conneci
    Connection::register('mysql', Custom\MySQL\Connection::class); // or directly using the class
    Custom\MySQL\Connection::register();
    
-The driver type used in the latter case is the default value of the `$driverType` property of `Custom\MySQL\Connection`
+The driver type used in the latter case is the default value of the `$driverType` property of 
+`Custom\MySQL\Connection`
 
-.. php:method:: createDriverConnection(array $dsn)
+.. php:method:: establishDriverConnection(array $dsn)
```
mvorisek: isn't `establish` a `connect`?

georgehristov: Yes, it is but Connection::establish is semantic.
So this word does the job - you understand that it connects 😄
I am also happy that it sounds good

mvorisek: `establish` is the same as `connect` to me, but `connect` sounds like simple english...

Anyway, I do not like `xxxDriverConnection` because it returns `Connection` and not a driver (like you claim in the name)...

georgehristov: It returns driver connection as it says in the name. Not a driver.
As I said connections in dsql are nested so this is a way to distinguish the nested connection which "drives" the parent.

mvorisek: I see - driver vs. connection. We should not mix it up, so probably only `establishDriver` here.

But still not sure if `establish` is the right term... I am not amazed with it, but not strictly negative...

src/Connection.php
mvorisek: @georgehristov but the driverType should be removed, it does not work with nested drivers... (to check if the underlaying driver is of some type)

As the check was not required until now, I would not introduce any helper method for it now until really required.

src/Connection.php
```diff
@@ -31,6 +31,22 @@ class Connection
...
+     * Stores the driverType => connectionClass array for resolving.
```
mvorisek: we should probably call driverType driver scheme (it is scheme when specified by URI, do more research on this how it is called internally in PHP/PDO lib)

tests/ConnectionTest.php
```diff
@@ -7,6 +7,26 @@
 use atk4\core\AtkPhpunit;
 use atk4\dsql\Connection;
 
+class DummyConnection extends Connection
```
mvorisek, line 10: drop with property

src/Connection.php
```diff
@@ -154,77 +154,72 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+     * MySQL\Connection::register()
+     *
+     * CustomDriver\Connection must be descendant of Connection class.
+     *
+     * @param string $driverType
+     * @param string $connectionClass
+     */
+    public static function registerDriver($driverType = null, $connectionClass = null)
+    {
+        if (!$connectionClass && is_a($driverType, self::class, true)) {
```
mvorisek: better to remove this and no default params...

never add this wild/bad designs

src/Connection.php
```diff
@@ -154,77 +154,72 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+    }
+
+    /**
+     * Resolves $dsn to a driver
+     * By default the driver is new PDO object which can be overridden in child classes.
      *
      * This does not silence PDO errors.
      */
-    protected static function getPDO(array $dsn)
+    public static function establishDriverConnection(array $dsn)
```
mvorisek: Why public?

src/Connection.php
```diff
@@ -104,43 +128,19 @@ public static function normalizeDSN($dsn, $user = null, $pass = null)
...
-                    // Default, for backwards compatibility
-                    $queryClass = Mysql\Query::class;
-
-                    break;
-            }
+
+            /**
+             * @var Connection $driverClass
+             */
+            $connectionClass = self::resolveDriver($driverType);
```
mvorisek: It does not resolve driver!

It does resolve connection class!

Driver is PDO, nothing else. There should be no driver elsewhere.

georgehristov: You need to dig a bit deeper in DSQL first.

I spent a lot of time last PR to figure out how to call the peoperties and classes as Connections are nested.
So it is connecton -> connection -> connecton -> PDO or connection or whatever that has `execute` method)

So finally now I decided to keep the old naming just call the wrapped connection "driverConnection".
`resolveDriver` gets a driver alias and resolves it to a connection class

mvorisek: I know, it is driver if not nested in execute.

But the new design should focus on proving strictly connection (you can still wrap it ;-) in your app if absolutely needed)

georgehristov: Yes, we can think of further redesign but this PRs is targeted at removing direct coupling of drivers in code and introducing the extendable registry

mvorisek: As long as you have this annotated to return a `Connection` (type should be added to the code anyway), than this name is 100% wrong. The same for all registry related names. The registry like factory for `Connection` based on `scheme`, right?

docs/connection.rst
romaninsh: since class is called Connection, calling method connect seems redundant. OK to rename into establish.

src/Connection.php
```diff
@@ -268,6 +263,16 @@ public function connection()
...
     }
 
+    /**
+     * Returns the default driver type set for the connection in $driverType.
+     *
+     * @return string|null
+     */
+    public static function defaultDriverType()
+    {
+        return (new \ReflectionClass(static::class))->getDefaultProperties()['driverType'] ?? null;
```
romaninsh, line 242: could you please make sure that the changes work with atk4/schema ?

docs/connection.rst
```diff
@@ -16,13 +16,13 @@ Connection class is handy to have if you plan on building and executing
 queries in your application. It's more appropriate to store
 connection in a global variable or global class::
 
-    $app->db = atk4\dsql\Connection::connect($dsn, $user, $pass);
+    $app->db = atk4\dsql\Connection::establish($dsn, $user, $pass);
```
romaninsh: shouldn't there be one more argument for args ?

docs/connection.rst
```diff
@@ -16,13 +16,13 @@ Connection class is handy to have if you plan on building and executing
...
 
 
-.. php:staticmethod:: connect($dsn, $user = null, $password = null, $args = [])
+.. php:staticmethod:: establish($dsn, $user = null, $password = null, $args = [])
```
romaninsh: one pitfall with passing password as an argument is that debugBacktrace could expose it. Could you think how to avoid the password being displayed if user have forgotten to turn off errors?

georgehristov:
@mvorisek I don't like the naming but I have reverted to using connect/Connection everywhere.
Feel free to modify further to what you feel correct and merge.

src/Connection.php
```diff
@@ -154,77 +149,72 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+     * MySQL\Connection::register()
+     *
+     * CustomDriver\Connection must be descendant of Connection class.
+     *
+     * @param string $connectionType
+     * @param string $connectionClass
+     */
+    public static function registerConnection($connectionType = null, $connectionClass = null)
+    {
+        if (!$connectionClass && is_a($connectionType, self::class, true)) {
```
mvorisek: This is very bad design. We should never swap any arguments. The method has some header with some params order and it should also work that way.

src/Connection.php
```diff
@@ -44,6 +60,9 @@ public function __construct($properties = [])
                 ->addMoreInfo('properties', $properties);
         }
 
+        // backward compatibility - to be removed with Connection::$driverType
+        $this->driverType = array_search(static::class, self::$connectionRegistry, true);
```
mvorisek: why is `$driverType` set also in Connection classes, can it be removed there?

src/Connection.php
```diff
@@ -154,77 +149,72 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
+    }
+
+    /**
+     * Resolves the connection class to use based on driver type.
+     *
+     * @param string $connectionType
+     *
+     * @return string
+     */
+    public static function resolveConnection($connectionType)
```
mvorisek: can we refactor whole registry from `Connection` to `ConnectionClass`?

src/Connection.php
mvorisek: this only duplicates registry defaults, I this we should not have this static method at all...

src/Connection.php
```diff
@@ -154,77 +149,72 @@ public static function connect($dsn, $user = null, $password = null, $args = [])
...
 
     /**
-     * Returns new PDO object.
+     * Establishes connection based on a $dsn
+     * By default connection is established using new PDO object which can be overridden in child classes.
      *
      * This does not silence PDO errors.
      */
-    protected static function getPDO(array $dsn)
+    public static function establishConnection(array $dsn)
```
mvorisek: If `connect` stays (and it should), this should never be public and what is the reason of the rename? We can name it `connectDriver()` (to be consistent with connect) and it should return `PDO` only.

The special classes should `connect` the underlaying drivers by overriding the `connect` method.

georgehristov: I will make it protected and rename it to connectDriver, this is a good name.
The reason for rename is that it does not always return PDO, t may also return another Connection.

src/Connection.php
georgehristov: We can `registry` designates what it is for. Can be also `connectionClassRegistry` if more clear

src/Connection.php
georgehristov: You are right, I was working on using classes only without `driverType` and this is left-over

src/Connection.php
mvorisek: fine with this

src/Connection.php
mvorisek: yes, much better

src/Connection.php
georgehristov: changed now

src/Connection.php
mvorisek: I actually liked this approach - do we need `defaultDriverType()` as all?

anyway, we can still implement `defaultDriverType()` like here and actually call that method here ;-)

and then we can remove all `public $driverType = 'xxx';`

The goal is to never specify driverType twice (once in registry, once in connection class).

src/Connection.php
mvorisek: you renamed only the properly, not related methods (`registerConnection`, `resolveConnection`)

src/Connection.php
mvorisek: image

but you introduces some more balast!

what is the realworld scenario where this provide any significant value for standard user?

if someone wants to register one class for more types, then it can be simply called twice!!!

src/Connection.php
georgehristov: At the moment the issue comes with nested connections (like ProxyConnection) where driverType comes from the nested connection.
The idea of having driverType is not changed from my code so better to plan such changes in another PR

src/Connection.php
georgehristov: It is like asking why we need e-mail if we can write letters on paper :)
It is a facilitation - but registering a list of connections removed now to keep things rudimentary.

src/Connection.php
mvorisek: Yes - I propose here to keep this removed line and remove all manual assigment. Once a connection class is registered, the effect will be the same. And no duplication.

src/Connection.php
mvorisek: rudimentary is perfect! :)

src/Connection.php
mvorisek: @georgehristov I went thru other feedback and this seems the last one to be addressed

mvorisek (approved):
seems fine, I think we can merge

PR #219, opened 2020-06-30
Add testing with PHP 8.0

mvorisek:


mvorisek:
@DarkSide666 wdyt? `_param` still accepts any scalar type (I think it is better so than casting it somewhere before)

tested/passing with `data` and `ui`

src/Expression.php
```diff
@@ -376,17 +358,12 @@ protected function _escape($value)
      * query rendering. Consider using `_consume()` instead, which will
      * also handle nested expressions properly.
      *
-     * @param string|array $value String literal or array of strings containing input data
+     * @param string|int|float $value
```
DarkSide666, line 361: I guess it should be `mixed` type here.

DarkSide666 (approved):
Looks fine.

src/Expression.php
mvorisek: mixed means any, I think it should be string or scalar (we should cast everything to string in dsql anyway...)

mvorisek: image

PR #220, opened 2020-07-01
Test with native MySQL 8.0

mvorisek:
`caching_sha2_password` auth plugin issue solved

see https://github.com/docker-library/mysql/issues/690

big, really big thank you to @tianon


DarkSide666 (approved):
wow - finally a solution. Thanks!

PR #226, opened 2020-07-07
Fix PSR class/trait/method naming

mvorisek:
related with https://github.com/atk4/core/pull/258

updated names:
```
Expressionable::getDSQLExpression()
Connection::lastInsertID()
Connection::normalizeDSN()
```
for 100% BC you can require the renamed classes manually (but if some other updated code required them before use is enough)


georgehristov (approved):
This will conflict with the driver registry PR...

mvorisek:
yes, I will update that PR - and hope we can merge it soon!

@georgehristov done

PR #231, opened 2020-07-31
Where does no longer accept array for OR conditions

mvorisek:
help wanted:
- doc must be update
- atk4/data must be updated, this test is failing:
  ```
  atk4\data\tests\ConditionSqlTest::testOrConditions
  ```

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -534,27 +534,7 @@ class Query extends Expression
      * surrounded by brackets:
      *  $q->where('user_id',$q->dsql()->table('users')->field('id'));
      *
-     * You can specify OR conditions by passing single argument - array:
-     *  $q->where([
-     *      ['a','is',null],
-     *      ['b','is',null]
-     *  ]);
-     *
-     * If entry of the OR condition is not an array, then it's assumed to
-     * be an expression;
-     *
-     *  $q->where([
-     *      ['age',20],
-     *      'age is null'
-     *  ]);
-     *
-     * The above use of OR conditions rely on orExpr() functionality. See
-     * that method for more information.
-     *
-     * To specify OR conditions
-     *  $q->where($q->orExpr()->where('a',1)->where('b',1));
-     *
-     * @param mixed  $field    Field, array for OR or Expression
+     * @param mixed  $field    Field or Expression
      * @param mixed  $cond     Condition such as '=', '>' or 'is not'
      * @param mixed  $value    Value. Will be quoted unless you pass expression
      * @param string $kind     Do not use directly. Use having()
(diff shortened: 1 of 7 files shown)


DarkSide666:
Why remove this functionality? Does it interferes with something?

mvorisek:
@DarkSide666 it is hack, we need to be sure we actually reject array/wrong input instead of threating it differently

As this method is more or less used only from data Scopes, we can easily make sure we do not generate this input

In the future, we may also introduce another junctions much more easiyl

@georgehristov promised to update the Scopes to not use this

mvorisek:
@georgehristov ok with you?

mvorisek:
@georgehristov can you please help with the docs for this?

mvorisek:
@georgehristov can you adjust data Scopes so we can merge this, eg. to not rely on array nesting level?

georgehristov:
@mvorisek so what is the solution for handling OR conditions now?
```
$q->where($q->orExpr()->where('a',1)->where('b',1));
```
Is it the one above? You have deleted the code comments for it

mvorisek:
yes, passing something like `$q->orExpr()` directly

georgehristov:
Prima-vista `data` does not rely on this "or from array" functionality but I am checking now

georgehristov:
Pls merge develop into this branch...

mvorisek:
rebased on develop

georgehristov:
Thanks. I just ran the tests on `data` using this branch and all is fine?

src/Query.php
```diff
@@ -534,27 +534,7 @@ public function _render_join()
...
-     *
-     *  $q->where([
-     *      ['age',20],
-     *      'age is null'
-     *  ]);
-     *
-     * The above use of OR conditions rely on orExpr() functionality. See
-     * that method for more information.
-     *
-     * To specify OR conditions
```
georgehristov, line 554: I would keep this below comment here

georgehristov (approved):

src/Query.php
mvorisek: 👌 , comment reverted

src/Query.php
```diff
@@ -571,17 +554,9 @@ public function where($field, $cond = null, $value = null, $kind = 'where', $num
...
-                if (is_array($row)) {
-                    $or->where(...$row);
-                } else {
-                    $or->where($row);
-                }
-            }
-            $field = $or;
+        // remove in v2.5
+        if (is_array($field)) {
+            throw new Exception('Array input / OR conditions is no longer supported');
```
DarkSide666, line 559: This was easy and handy method how to define OR conditions :(

mvorisek: and how to possibly and silently pass mistakes

PR #232, opened 2020-07-31
Fix doc for Model::action to Model::toQuery refactor

mvorisek:
merge after https://github.com/atk4/data/pull/688


mvorisek:
we can reopen later or this repo may be merged with data

PR #233, opened 2020-08-03
[update] convert Expressionable value in Query::_sub_render_condition

georgehristov:
Related to https://github.com/atk4/data/pull/690

### DIFF
--- a/src/Query.php
+++ b/src/Query.php
@@ -770,6 +770,10 @@ class Query extends Expression
         if (count($row) === 2) {
             $value = $cond;
 
+            if ($value instanceof Expressionable) {
+                $value = $value->getDsqlExpression($this);
+            }
+
             if (is_array($value)) {
                 $cond = 'in';
             } elseif ($value instanceof self && $value->mode === 'select') {


src/Query.php
```diff
@@ -770,6 +770,10 @@ protected function _sub_render_condition($row)
         if (count($row) === 2) {
             $value = $cond;
 
+            if ($value instanceof Expressionable) {
```
mvorisek, line 773: where it is needed? only for count() === 2?

georgehristov: Yes, it is intended to uncover a nested Query and resolve correctly a few lines below in `instance of self`
In atk4/data (revamped) we will have `Persistence\Sql\Query implements Expressionable`. On getDsqlExpression it returns the underlying dsql\Query.
This change here is intended to uncover the dsql\Query so the operator is resolved correctly to 'in' and not '='

src/Query.php
```diff
@@ -770,6 +770,10 @@ protected function _sub_render_condition($row)
         if (count($row) === 2) {
             $value = $cond;
```
mvorisek, line 771: we should definitely parse to `$value` directly

src/Query.php
mvorisek: A little cryptic - but if this will be covered by some test/common usecase (here on in data) - then ok

enforce return type of `dsql\Expression` from all `getDsqlExpression` methods

mvorisek (approved):

PR #236, opened 2020-09-02
Fix CI PostreSQL errors

mvorisek:


mvorisek:
issue in data repo...

PR #238, opened 2020-09-05
Fix "is null" for PostgresSQL

mvorisek:
PostgresSQL does parse `expr is null` only (instead of ` expr is $1`, where `$1` represents a bounded variable)


georgehristov (approved):
LGTM

PR #246, opened 2020-09-09
Fix tests for MSSQL

mvorisek:


src/Expression.php
```diff
@@ -542,10 +538,13 @@ public function execute(object $connection = null)
                 $statement->execute();
             } catch (\PDOException $e) {
                 $new = (new ExecuteException('DSQL got Exception when executing this query', $e->errorInfo[1]))
-                    ->addMoreInfo('error', $e->errorInfo[2])
+                    ->addMoreInfo('error', $e->errorInfo[2] ?? 'n/a (' . $e->errorInfo[0] . ')')
```
mvorisek, line 541: `errorInfo[2]` is not set with MSSQL PDO driver

src/Expression.php
```diff
@@ -522,13 +522,9 @@ public function execute(object $connection = null)
...
 
-                    // workaround to support LOB data type 2/2, see https://github.com/doctrine/dbal/pull/2434
+                    // workaround to support LOB data type 2/3, see https://github.com/doctrine/dbal/pull/2434
                     $statement->boundValues[$key] = $val;
-                    if ($type === \PDO::PARAM_STR) {
-                        $bind = $statement->bindParam($key, $statement->boundValues[$key], $type, strlen((string) $val));
-                    } else {
-                        $bind = $statement->bindParam($key, $statement->boundValues[$key], $type);
-                    }
+                    $bind = $statement->bindParam($key, $statement->boundValues[$key], $type);
```
mvorisek, line 527: see https://social.msdn.microsoft.com/Forums/biztalk/en-US/c7431849-8594-4da2-bef6-8702b72391cc/php-pdo-bindparam-optional-length-parameter-problem#answers

lenght should be set only if param is read back (in-out), not our case

mvorisek:
@georgehristov can we merge?

georgehristov (approved):
LGTM

PR #247, opened 2020-09-13
Fix MSSQL Expression::execute() for nested calls 

georgehristov:
`Expression::$params` must always be array, yet in `atk4\dsql\Mssql\ExpressionTrait::getDebugQuery` it was assigned `paramsBackup` which could be null

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -500,6 +500,8 @@ class Expression implements \ArrayAccess, \IteratorAggregate
 
             $query = $this->render();
 
+            $statement = null;
+
             try {
                 $statement = $connection->prepare($query);
 
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -500,8 +500,6 @@ class Expression implements \ArrayAccess, \IteratorAggregate
 
             $query = $this->render();
 
-            $statement = null;
-
             try {
                 $statement = $connection->prepare($query);
 
--- a/src/Mssql/ExpressionTrait.php
+++ b/src/Mssql/ExpressionTrait.php
@@ -23,7 +23,7 @@ trait ExpressionTrait
 
     // {{{ MSSQL does not support named parameters, so convert them to numerical inside execute
 
-    private $paramsBackup = [];
+    private $paramsBackup;
     private $fixedRender;
 
     public function execute(object $connection = null)
@@ -47,6 +47,7 @@ trait ExpressionTrait
             return parent::execute($connection);
         } finally {
             $this->params = $this->paramsBackup;
+            $this->paramsBackup = null;
             $this->fixedRender = null;
         }
     }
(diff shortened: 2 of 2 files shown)


mvorisek (approved):
I think I coded it wrongly, please double check before merge

src/Expression.php
```diff
@@ -500,6 +500,8 @@ public function execute(object $connection = null)
 
             $query = $this->render();
 
+            $statement = null;
```
mvorisek: absolutely not needed

src/Mssql/ExpressionTrait.php
```diff
@@ -23,7 +23,7 @@ protected function _escape(string $value): string
 
     // {{{ MSSQL does not support named parameters, so convert them to numerical inside execute
 
-    private $paramsBackup;
+    private $paramsBackup = [];
```
mvorisek: it seems my fix is not working, but this is WRONG - we either need to use a backup or not

backup it there to support named parameters and convert them to a numerical (`:param` -> `?`) for MSSQL, as MSSQL does not support named parameters

it seems, that something is called twice and thus backup thru a shared/object variable does not work - fix this, or implement native support to convert params before query in Expression::execute method

georgehristov:
@mvorisek it still triggers the same error. I do not have Mssql to test it so have a look if you want it fixed your way

mvorisek:
> 
> 
> @mvorisek it still triggers the same error. I do not have Mssql to test it so have a look if you want it fixed your way

see above, you did not fix this issue, you only suppressed some type error somewhere :)

where did you spot this issue?

georgehristov:
The issue comes when you call `atk4\dsql\Mssql\Expression::getDebugQuery`.

please post exact steps to reproduce, I will into it

georgehristov:
> Revert "[fix] set $statement"

What happens if the exception is within `$connection->prepare($query)` and $statement is not initialized (in `Expression::execute`)?

mvorisek:
> 
> 
> > Revert "[fix] set $statement"
> 
> What happens if the exception is within `$connection->prepare($query)` and $statement is not initialized (in `Expression::execute`)?

how does your fix help?

I belive:
1. PDO will now throw anything else than `PDOException`
2. `$statement = null;` does not help in any sense - we should never return null :)

georgehristov:
Then you need to wrap `unset($statement->{'boundValues'});` in `if (isset($statement))`.
If the exception is within `$connection->prepare($query)` `$statement` is not initialized

mvorisek:
> 
> 
> Then you need to wrap `unset($statement->{'boundValues'});` in `if (isset($statement))`.
> If the exception is within `$connection->prepare($query)` `$statement` is not initialized

your are right with this, I commited a proper fix for it

georgehristov:
But the original issue still remains...

georgehristov:
RTM?

mvorisek:
yes...

georgehristov:
Thanks!

PR #248, opened 2020-09-13
Expression implements Expressionable

georgehristov:


src/Expression.php
```diff
@@ -8,7 +8,7 @@
  * Creates new expression. Optionally specify a string - a piece
  * of SQL code that will become expression template and arguments.
  */
-class Expression implements \ArrayAccess, \IteratorAggregate
+class Expression implements Expressionable, \ArrayAccess, \IteratorAggregate
```
mvorisek: `Expressionable` interface should imply a class with basic expression methods, not a getter for an expression

src/Expression.php
```diff
@@ -500,6 +505,8 @@ public function execute(object $connection = null)
 
             $query = $this->render();
 
+            $statement = null;
```
mvorisek: solved in https://github.com/atk4/dsql/pull/249 , revert

mvorisek (changes requested):
Logical from names, ilogical from the design - why should "expressionable expression provide a method, to get an expression"?

mvorisek:
if not needed for some larger refactoring, I would close this one.

mvorisek:
@georgehristov how much do we need this?

I am still not sure about this change as allowing `Expressionable` everywhere would complicate static analysis, basically everywhere where we would annotate `Expressionable` (instead of `Expression`) we would have to use `getDsqlExpression`

So I propose to actually drop `Expressionable` at all to `IExpression` that will define minimal expression interface which other non `Expression` classes (like data `Field`) can implement.

mvorisek (approved):
well, why not, thanks @georgehristov

PR #249, opened 2020-09-13
Remove bound values workaround for LOB datatype

mvorisek:
see "experiment with unset before execute" commit - for some reasons, plain variable can not be passed, but array element can be passed and this array can be unset even before `PDOStatement::execute()`


georgehristov:
failing tests?

mvorisek:
> 
> 
> failing tests?

I see, fixed...

now I done more tests and it seems that the `$boundValues` variable can be unset even before execute

mvorisek:
@georgehristov with bindValue, there should be no issue even with Oracle, let's merge this

georgehristov (approved):

PR #250, opened 2020-09-15
Introduce connection specific exists query

georgehristov:


tests/OracleTest.php
```diff
@@ -114,4 +114,12 @@ public function test12OracleLimitSkip()
             $c->dsql()->table('foo')->where('bar', 1)->field('baz')->limit(10, 99)->render()
         );
     }
+
+    public function testOracleExists()
```
mvorisek, line 118: `OracleTest.php` file/tests will be dropped soon Oracle DB will be fully tested in CI, just FYI, we can merge now...

mvorisek (approved):
very good

tests/OracleTest.php
georgehristov: Then in testExists we should check for each connection? It is already complaining that Mssql\Query::exists is not tested...

mvorisek: yes (and I expect that it is already tested indirectly in atk4/data)

PR #261, opened 2020-10-14
Drop driverType property and debug connection classes

mvorisek:
merge once data/schema PR is ready

BC break because:

- removed `driverType` property
- DBAL `XxxPlatform::getName()` is different vs. the original `driverType` property for some vendors, see https://github.com/doctrine/dbal/tree/v2.9.3/lib/Doctrine/DBAL/Platforms , ie.:
  - `sqlite` - no change
  - `mysql` - no change
  - `sqlsrv` -> `mssql`
  - `pgsql` -> `postgresql`
  - `oci` -> `oracle`
- removed debug connection classes
- removed deprecated connection classes
- DSN `driverType` key renamed to `driverSchema`
- removed `Connection::defaultDriverType` static method


docs/advanced.rst
```diff
@@ -75,53 +75,6 @@ to explicitly specify property :php:attr:`Connection::query_class`::
 This is also useful, if you have created your own Query class in a different
 namespace and wish to use it.
 
-Using Debug Proxy Connections (Stopwatch and Profiler)
```
georgehristov, line 78: Why are we dropping Stopwatch and Profiler proxy connections?

mvorisek: Because they are wrongly designed:
- if driverType is/was used somewhere, it was meant to be used to check the underlaying DB type, ie. with debug connections it will not work
- in the future, we should replace connections classes with DBAL connection classes which support debug hooks which is the correct way to achive that debug support (now, you can implement it in a few lines on your own)

we should remove Proxy connection as well and simplify dsql even more (likie nested execute - we should not really care about it)

georgehristov: I am fine with this as long as there is way to debug

mvorisek: please approve if ok, RTM, I will create now a PR for data to adjust the code

keeping ProxyConnection for now

georgehristov (approved):
LGTM

PR #262, opened 2020-10-17
Use DBAL connection instead of PDO directly

### LINKED ISSUE #223: Replace Connection with DBAL
I would highly prefer to drop all Connection classes in advance to DBAL. This will not solve only the registry, but it would allow to share that connection with Doctrine ORM much easier later.

How Doctrine setup connection?
doc - https://www.doctrine-project.org/projects/doctrine-orm/en/2.7/tutorials/getting-started.html#obtaining-the-entitymanager
which refer to https://github.com/doctrine/orm/blob/1ab9d0b1942a6b70d01640d51056a71916255ecf/lib/Doctrine/ORM/EntityManager.php#L825
which calls https://github.com/doctrine/orm/blob/2a98a98cd36e5eb54ad5ff8ca26fc3a30d5a82ff/lib/Doctrine/ORM/EntityManager.php#L842
which calls a few lines later https://github.com/doctrine/orm/blob/2a98a98cd36e5eb54ad5ff8ca26fc3a30d5a82ff/lib/Doctrine/ORM/EntityManager.php#L845
DriverManager is part of DBAL.

mvorisek:
fixes #223

we need it if we want to use DBAL Schema Manager

BC break:

- `Connection::connection()` method now returns DBAL Connection instead of PDO
- to get PDO (or PDO like driver), call `->getWrappedConnection()` on it
- drop proxy connection class (you should use DBAL hooks)


src/Connection.php
```diff
@@ -214,15 +225,13 @@ public function expr($properties = [], $arguments = null): Expression
     {
         $c = $this->expression_class;
         $e = new $c($properties, $arguments);
-        $e->connection = $this->connection ?: $this;
+        $e->connection = $this;
```
mvorisek, line 225: this is original design inconsistency

src/Connection.php
```diff
@@ -364,7 +373,7 @@ public function rollBack()
      */
     public function lastInsertId(string $sequence = null): string
     {
-        return $sequence === null ? $this->connection()->lastInsertId() : $this->connection()->lastInsertId($sequence);
+        return $this->connection()->lastInsertId($sequence);
```
mvorisek, line 343: null can be passed, tested

src/Expression.php
```diff
@@ -495,8 +496,8 @@ public function execute(object $connection = null)
         }
 
         // If it's a PDO connection, we're cool
-        if ($connection instanceof \PDO) {
-            $connection->setAttribute(\PDO::ATTR_ERRMODE, \PDO::ERRMODE_EXCEPTION);
+        if ($connection instanceof DbalConnection) {
+            $connection = $connection->getWrappedConnection();
```
mvorisek, line 502: ERRMODE_EXCEPTION is set by DBAL connection on init

DarkSide666 (approved):
LGTM

PR #263, opened 2020-10-18
Refactor to not use reference properties

### LINKED ISSUE #200: Remove reference property
Refactor the current code to not use property references as PHP references are very fragile to errors.

See https://github.com/FriendsOfPHP/PHP-CS-Fixer/issues/4939

All major repos - core, schema, data, ui are ok expect this repo, the references are set at:
- https://github.com/atk4/dsql/blob/7f6ff3d9c8479f42674488b8bb4b31c86d022d22/src/Expression.php#L282-L283 and unset at:
- https://github.com/atk4/dsql/blob/7f6ff3d9c8479f42674488b8bb4b31c86d022d22/src/Expression.php#L293
- https://github.com/atk4/dsql/blob/7f6ff3d9c8479f42674488b8bb4b31c86d022d22/src/Expression.php#L463

mvorisek:
fixes #200

tests are passing, please review carefully

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -50,7 +50,7 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      *
      * @var string
      */
-    public $_paramBase;
+    private $_paramBase;
 
     /**
      * Will be populated with actual values by _param().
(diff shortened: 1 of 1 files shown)


romaninsh:
I no longer recall why references were used here. But I'm glad if they can be removed.

mvorisek:
> 
> 
> I no longer recall why references were used here. But I'm glad if they can be removed.

lazyness of some programmer... it is internal & original hack for `_paramBase` only

DarkSide666 (approved):
LGTM

PR #264, opened 2020-10-18
Do not escape/template in quoted SQL string literals/identifiers

### LINKED ISSUE #144: Escaping in expressions

mvorisek:
fixes https://github.com/atk4/dsql/issues/144

string grammars:
- https://github.com/antlr/grammars-v4/blob/729c0a5ae0a22b3f833a906ba45ae5d394252a08/sql/sqlite/SQLiteLexer.g4#L224
- https://github.com/antlr/grammars-v4/blob/a226f52cbf4e1bec86647a9a9dbabda3db8cbe5b/sql/mysql/Positive-Technologies/MySqlLexer.g4#L1236
- https://github.com/antlr/grammars-v4/blob/04fd2b529eb6b4db4944dc7ba96bbb596eefb155/sql/tsql/TSqlLexer.g4#L839
- https://github.com/antlr/grammars-v4/blob/cbc94950e81ffb62badcaf8e00ab8cabc3e1d3dd/sql/plsql/PlSqlLexer.g4#L2283

identifier grammars:
- https://github.com/antlr/grammars-v4/blob/729c0a5ae0a22b3f833a906ba45ae5d394252a08/sql/sqlite/SQLiteLexer.g4#L213
- https://github.com/antlr/grammars-v4/blob/a226f52cbf4e1bec86647a9a9dbabda3db8cbe5b/sql/mysql/Positive-Technologies/MySqlLexer.g4#L1199
- https://github.com/antlr/grammars-v4/blob/04fd2b529eb6b4db4944dc7ba96bbb596eefb155/sql/tsql/TSqlLexer.g4#L831
- https://github.com/antlr/grammars-v4/blob/cbc94950e81ffb62badcaf8e00ab8cabc3e1d3dd/sql/plsql/PlSqlLexer.g4#L2297

BC break:
yes, but uncommon, only if you use templates like `` '`{{x}}`' `` (escape using templating, ie. `'{x}'`), see one problematic test

### DIFF
--- a/tests/ExpressionTest.php
+++ b/tests/ExpressionTest.php
@@ -210,9 +210,7 @@ class ExpressionTest extends AtkPhpunit\TestCase
                 '\'\'\'[]\'',
                 '\'[]\'\'\'',
             ] as $testStr) {
-                if ($enclosureChar !== '\'') {
-                    $testStr = str_replace('\'', $enclosureChar, $testStr);
-                }
+                $testStr = str_replace('\'', $enclosureChar, $testStr);
 
                 yield [$testStr, $testStr, []];
 
--- a/tests/ExpressionTest.php
+++ b/tests/ExpressionTest.php
@@ -188,6 +188,49 @@ class ExpressionTest extends AtkPhpunit\TestCase
     }
 
     /**
+     * @dataProvider provideNoTemplatingInSqlStringData
+     */
+    public function testNoTemplatingInSqlString(string $expectedStr, string $exprStr, array $exprArgs)
+    {
+        $this->assertSame($expectedStr, $this->e($exprStr, $exprArgs)->render());
+    }
+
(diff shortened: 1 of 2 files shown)


tests/ExpressionTest.php
```diff
@@ -164,7 +164,7 @@ public function testConstructor4()
         $this->assertSame(
             'testing "hello, world"',
             $this->e(
-                'testing "[]"',
+                'testing []',
```
mvorisek, line 167: @georgehristov @DarkSide666 how would like you to see this fixed? - remove the test completely?

romaninsh (approved):
Looks good

PR #266, opened 2020-10-24
Query everything thru DBAL connection (to trigger DBAL hooks)

mvorisek:
here is example how to hook for all executed queries:
https://github.com/atk4/data/blob/5a67e085474005c8218b345b224e8f1bb2d94098/src-schema/PhpunitTestCase.php#L42-L53

aka replace ConnectionDumper or how it was called...

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace atk4\dsql;
 
 use Doctrine\DBAL\Connection as DbalConnection;
+use Doctrine\DBAL\DBALException;
 
 /**
  * Creates new expression. Optionally specify a string - a piece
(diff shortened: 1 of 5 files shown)


georgehristov (approved):
LGTM

src/Expression.php
```diff
@@ -522,8 +523,6 @@ public function execute(object $connection = null)
 
         // If it's a PDO connection, we're cool
         if ($connection instanceof DbalConnection) {
-            $connection = $connection->getWrappedConnection();
```
mvorisek, line 525: this is basically the one and only MAJOR change

georgehristov:
Working on running everything on DBAL, keeping only Expression class as ATK specific and very good feature

mvorisek:
Yes, dropping DSQL/Connection mess & specific DB vendor implementations is highly welcomed, then DSQL/Expression and DSQL/Query (too) should probably stay for the first PRs.

I think, there should be two PRs - one for dropping all DSQL DB vendor classes, one for dropping the actual DSQL Connection class then.

PR #268, opened 2020-10-29
[feature] introduce Expression escape constants

georgehristov:


src/Expression.php
```diff
@@ -13,6 +13,11 @@
  */
 class Expression implements \ArrayAccess, \IteratorAggregate
 {
+    public const ESCAPE_PARAM = 'param';
```
mvorisek: This is internal only, can it be protected?

Also annotate, even I do not like uneeded docs in general, like:
`/** @const string "[]" in template, escape as string */`
`/** @const string "{}" in template, escape as identifier */`
`/** @const string "{{}}" in template, escape as identifier, but keep input with special characters like "." or "(" unescaped */`

mvorisek: I think it should be more like PARAM, IDENTIFIER, IDENTIFIER_SOFT, NONE

mvorisek (changes requested):

mvorisek:
As protected (ok, can be possibly used somewhere in extended classes), I think the names MUST match the constant values. Very important to prevent human conflicts.

georgehristov:
Yes, the names will match the constants and now they do.
If we are to change them the should be:
PARAM = PARAM
COMPLETE = STRING
SOFT = STRING_SOFT
NONE = NONE

mvorisek:
> 
> 
> Yes, the names will match the constants and now they do.
> If we are to change them the should be:
> PARAM = PARAM
> COMPLETE = STRING
> SOFT = STRING_SOFT
> NONE = NONE

no!

STRING (literal) === param
IDENTIFIR === name

as STRING is set using bound values, yes, then PARAM is ok/good

methods should be probabaly renamed as well, as `Expr::escape()` is too unspecific name for identifier escaping/quoting

georgehristov:
Methods is next phase - should be easy as they are protected.

Let me know your complete name mapping suggestion (CURRENT > NEW)

mvorisek:
> 
> 
> Methods is next phase - should be easy as they are protected.
> 
> Let me know your complete name mapping suggestion (CURRENT > NEW)

see updated https://github.com/atk4/dsql/pull/268#discussion_r514230598

mvorisek (approved):
seems good to me

src/Mssql/ExpressionTrait.php
```diff
@@ -6,19 +6,19 @@
...
 
-    protected function _escapeSoft(string $value): string
+    protected function escapeIdentifierSoft(string $value): string
     {
-        return $this->fixOpenEscapeChar(parent::_escapeSoft($value));
+        return $this->fixOpenEscapeChar(parent::escapeIdentifierSoft($value));
     }
 
-    protected function _escape(string $value): string
+    private function fixOpenEscapeChar(string $v)
```
mvorisek:
```suggestion
    private function fixOpenEscapeChar(string $v): string
```

src/Query.php
```diff
@@ -814,7 +814,7 @@ protected function _sub_render_condition($row)
                 return '1 = 1'; // always true
             }
 
-            $value = '(' . implode(',', array_map([$this, '_param'], $value)) . ')';
+            $value = '(' . implode(',', array_map([$this, 'escapeParam'], $value)) . ')';
```
mvorisek:
```suggestion
            $value = '(' . implode(',', array_map(function ($v) { return $this->escapeParam($v); }, $value)) . ')';
```

mvorisek (approved):
👍

PR #269, opened 2020-10-29
[fix] wrap where expression in parenthesis when consumed

### LINKED ISSUE #239: No wrapping parenthesis around nested conditions
Just a small hint for you out there - if you do $model->addCondition('somefield', somvalue); and another $model->addCondition($model->expr('otherfield=1 or otherfield=2'). ATK4 will translate it into 'somefield=somevalue and otherfield=1 or otherfield=2'). In the logic of adding conditions and coupling them with AND this is not right. It should be 'somefield=somevalue and (otherfield=1 or otherfield=2)'.

So you need to put brackets yourself in the ->expr('(otherfield=1 or otherfield=2)') to get the desired result...

reported by @mkrecek234, thanks

slightly related with https://github.com/atk4/dsql/pull/231

georgehristov:
Solves https://github.com/atk4/dsql/issues/239

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -78,11 +78,11 @@ class Expression implements \ArrayAccess, \IteratorAggregate
     public $connection;
 
     /**
-     * Wrap the expression in parenthesis when consumed by another expression or not.
+     * Wrap the expression in parentheses when consumed by another expression or not.
      *
      * @var bool
      * */
-    public $consumeWrappedInParenthesis = false;
+    public $wrapInParentheses = false;
 
     /**
      * Specifying options to constructors will override default
@@ -296,13 +296,13 @@ class Expression implements \ArrayAccess, \IteratorAggregate
         }
 
         if (isset($expression->allowToWrapInParenthesis)) {
-            'trigger_error'('Usage of Query::$allowToWrapInParenthesis is deprecated, will be removed in version 2.5', E_USER_DEPRECATED);
+            'trigger_error'('Usage of Query::$allowToWrapInParenthesis is deprecated, use $wrapInParentheses instead - will be removed in version 2.5', E_USER_DEPRECATED);
 
-            $expression->consumeWrappedInParenthesis = $expression->allowToWrapInParenthesis;
+            $expression->wrapInParentheses = $expression->allowToWrapInParenthesis;
         }
 
         // Wrap in parentheses if expression requires so
-        if ($expression->consumeWrappedInParenthesis === true) {
+        if ($expression->wrapInParentheses === true) {
             $ret = '(' . $ret . ')';
         }
 
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -77,7 +77,11 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      */
     public $connection;
 
-    /** @var bool Do we wrap the expression in parenthesis when rendering? */
+    /**
+     * Wrap the expression in parenthesis when consumed by another expression or not.
+     *
+     * @var bool
+     * */
     public $consumeWrappedInParenthesis = false;
 
     /**
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -77,6 +77,9 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      */
     public $connection;
 
+    /** @var bool Do we wrap the expression in parenthesis when rendering? */
+    public $consumeWrappedInParenthesis = false;
+
     /**
      * Specifying options to constructors will override default
      * attribute values of this class.
@@ -262,8 +265,8 @@ class Expression implements \ArrayAccess, \IteratorAggregate
                     return $expression;
             }
 
-            throw (new Exception('$escape_mode value is incorrect'))
-                ->addMoreInfo('escape_mode', $escapeMode);
+            throw (new Exception('$escapeMode value is incorrect'))
+                ->addMoreInfo('escapeMode', $escapeMode);
         }
 
         // User may add Expressionable trait to any class, then pass it's objects
@@ -288,8 +291,14 @@ class Expression implements \ArrayAccess, \IteratorAggregate
             $expression->_paramBase = null;
         }
 
-        // Queries should be wrapped in parentheses in most cases
-        if ($expression instanceof Query && $expression->allowToWrapInParenthesis === true) {
+        if (isset($expression->allowToWrapInParenthesis)) {
+            'trigger_error'('Usage of Query::$allowToWrapInParenthesis is deprecated, will be removed in version 2.5', E_USER_DEPRECATED);
+
+            $expression->consumeWrappedInParenthesis = $expression->allowToWrapInParenthesis;
+        }
+
+        // Wrap in parentheses if expression requires so
+        if ($expression->consumeWrappedInParenthesis === true) {
             $ret = '(' . $ret . ')';
         }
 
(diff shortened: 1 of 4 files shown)


src/Query.php
```diff
@@ -674,6 +670,12 @@ public function where($field, $cond = null, $value = null, $kind = 'where', $num
 
         switch ($num_args) {
             case 1:
+                if (is_string($field)) {
+                    $field = $this->expr($field);
+                }
+
+                $field->consumeWrappedInParenthesis = true;
```
georgehristov: All expressions are wrapped on principle better safe than sorry.
It can be precised by checking if expression template contains and/or words but `Expression::$template` is protected

src/Expression.php
```diff
@@ -77,6 +77,13 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      */
     public $connection;
 
+    /**
+     * Wrap the expression in parenthesis when consumed by another expression or not.
```
mvorisek:
```suggestion
     * Wrap in parenthesis when consumed by another expression.
```

src/Expression.php
```diff
@@ -77,6 +77,13 @@ class Expression implements \ArrayAccess, \IteratorAggregate
...
+     *
+     * @var bool
+     * */
+    public $consumeWrappedInParenthesis = false;
```
mvorisek: why past tense?

src/Query.php
mvorisek: I belive this is good, but is only in where enough?

what if we define field x = `a, b` expr, when we select it, will it be wrapped in select like `select (a, b) as x` ?

mvorisek: in other cases/num_args != 1, there is no issue?

src/Expression.php
georgehristov: This is how you say it as an adjective: something needs to be wrapped or not. "Wrap" is the verb - the act of doing so

mvorisek: yes, what grades did you have?

somethings needs to be wrapped, yes

consumeWraps - when expr is beiing consumed, consume wraps

PS: not a native speaker, but always A+ in Czech :D

georgehristov: No, I wrote it in English, not in Czech :)

It is written correctly in English.

It is meant not what is done to it (wrap as verb) but how it is consumed (wrapped as adjective).
The verb is "consume".
"Consume the expression wrapped in paranthesis."

mvorisek: You mean "it is wrapped", I meant "consume wraps" - when it arrives to consume, is it already wrapped?

If yes, I think you are right. If no, and concume does it, then I think, wrapped is not correct.

georgehristov: I understand you mean "consume method wraps it" but this is not an indicator of what the method has done to it but how it should be consumed.
"Consumption" is performed by the method.
So the method checks if expression should be consumed wrapped or not and performs accordingly.

mvorisek: I got it - consume once wrapped :))

but I find it confusing, as we asks consume to do it so

that is why named `allow...` previously... 3rd person feedback would be welcomed, @DarkSide666, would it be yours?

georgehristov: I think "allow" is not intuitive as the wrapping is strictly determined only by this setting. Allow implies it is optional - it can be allowed but not done, which is not the case. 
Let's see what @DarkSide666 will say.

mvorisek: There was something about UNION when it was not allowed to wrap if I remember...

georgehristov: Yes, queries are wrapped by default when consumed. But when union is used this can be switched off as some platforms don't support wrapped union queries.
Now I added that where clauses from an expression are also wrapped in paranthesis.

Let's merge this and move on to next issue...

mvorisek (changes requested):
What alternatives do you propose instead of consumeWrappedSomething? - I personally find this name VERY confusing, I am not gooing to approve this PR with this this name.

georgehristov:
`$wrapInParenthesesWhenConsumed` or better `$wrapInParentheses`

mvorisek:
> 
> 
> `$wrapInParenthesesWhenConsumed` or better `$wrapInParentheses`

I think `allow` is still slighly better as this is allowance for foreign operation (at least was when it was in Query class), but both proposed variants are ok with me, the main confusion was solved. 👍 

About the fix - is fix for the case with 1 arg only 100% guaranteed to be enough?

tests/QueryTest.php
```diff
@@ -824,6 +824,14 @@ public function testWhereBasic()
         );
     }
 
+    public function testWhereExpression()
```
mvorisek, line 827: Please add functional test (test testing the issue with real DB)

src/Expression.php
```diff
@@ -77,6 +77,13 @@ class Expression implements \ArrayAccess, \IteratorAggregate
      */
     public $connection;
 
+    /**
+     * Wrap the expression in parentheses when consumed by another expression or not.
+     *
+     * @var bool
+     * */
+    public $wrapInParentheses = false;
```
mvorisek, line 85: I think, if declared in Expression, we can make in protected and no need for BC at all

georgehristov: Not really as we need access externally. See below:

https://github.com/atk4/dsql/blob/519410d9fb52fc238eab66dc1117a1f84230b4bf/tests/QueryTest.php#L684-L693

georgehristov:
> About the fix - is fix for the case with 1 arg only 100% guaranteed to be enough?

The issue was when adding pure SQL containing 'and' / 'or' which is only done using single argument to `Query::where` so I believe it is solved.
If you can figure out a case where you can add it with more args then let me know.

src/Expression.php
mvorisek: you are right

tests/QueryTest.php
georgehristov: If SQL is rendering correctly I see no point in testing on real DB.
I updated the test a bit with more conditions

mvorisek:
> 
> 
> > About the fix - is fix for the case with 1 arg only 100% guaranteed to be enough?
> 
> The issue was when adding pure SQL containing 'and' / 'or' which is only done using single argument to `Query::where` so I believe it is solved.
> If you can figure out a case where you can add it with more args then let me know.

it is only theoretical, as you can still pass expr...

tests/QueryTest.php
mvorisek: there is - real DB tests a behaviour, rendered conditions can be very easily replaced when modifying something like now

mvorisek (approved):
I cleaned it a little bit, main point is addressed. I would still like to see one simple functional test testing the real and very major issue. Why explained in https://github.com/atk4/dsql/pull/269#discussion_r514978163

georgehristov:
> I would still like to see one simple functional test testing the real and very major issue

Added, merging this

src/Expression.php
DarkSide666: In case of union it has to be `SELECT * FROM t1 UNION SELECT * FROM t2` without wrapping select queries in parenthesis. I don't remember exactly now which engine didn't support `(select ...) union (select ...)`, but it was Oracle or PostGRE i guess.

georgehristov: It is Sqlite as per code comments
Now we use this same feature to wrap expressions used as where conditions in ()
Currently all conditions passed as expressions will be wrapped

PR #271, opened 2020-10-30
Rename Expression::get() to Expression::getRows()

mvorisek:
unify naming, prevent confusion


src/Expression.php
```diff
@@ -630,7 +630,7 @@ private function getCastValue($v): ?string
      *
      * @return string[][]|null[][]
      */
-    public function get(): array
+    public function getRows(): array
```
georgehristov, line 643: We need to have the old method issuing deprecated warning for 1 version iteration

georgehristov (approved):

PR #273, opened 2020-10-31
Fix Oracle CLOB/BLOB SQL limitation

mvorisek:
needed for unified DB support for atk4/data and possibly atk4/ui

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -7,6 +7,7 @@ namespace atk4\dsql;
 use Doctrine\DBAL\Connection as DbalConnection;
 use Doctrine\DBAL\DriverManager;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\OraclePlatform;
 
 /**
  * Class for establishing and maintaining connection with your database.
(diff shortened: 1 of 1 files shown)


src/Connection.php
```diff
@@ -196,7 +197,45 @@ protected static function connectDbalConnection(array $dsn)
...
+        if ($connection->getDatabasePlatform() instanceof OraclePlatform) {
+            \Closure::bind(function () use ($connection) {
+                $connection->platform = new class() extends OraclePlatform {
+                    private function forwardTypeDeclarationSQL(string $targetMethodName, array $column): string
+                    {
+                        $backtrace = debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT | DEBUG_BACKTRACE_IGNORE_ARGS);
+                        foreach ($backtrace as $frame) {
+                            if ($this === ($frame['object'] ?? null)
+                                && $targetMethodName === ($frame['function'] ?? null)) {
+                                throw new Exception('Long CLOB/TEXT (4000+ bytes) is not supported for Oracle');
```
DarkSide666, line 214: I still don't see the reason why it's not supported. CLOB definitely can be bigger than 4000 chars.
And what about BLOB?

mvorisek: Because https://stackoverflow.com/questions/12980038/ora-00932-inconsistent-datatypes-expected-got-clob#12980560, Oracle simply does not allow CLOB/BLOB types to be used in SQL directly, even if bound inside queiries like `UPDATE t SET x = []`.

DarkSide666 (approved):
OK for now

PR #276, opened 2020-11-30
Add PHPStan to CI

mvorisek:


DarkSide666 (approved):
Approved, but should fix these errors in phpdoc comments before merging

src/Query.php
```diff
@@ -709,7 +711,7 @@ protected function _sub_render_condition($row)
 
         // if no condition defined - set default condition
         if (count($row) === 2) {
-            $value = $cond;
+            $value = $cond; // @phpstan-ignore-line see https://github.com/phpstan/phpstan/issues/4173
```
mvorisek, line 714: related with https://github.com/phpstan/phpstan/issues/417

PR #281, opened 2020-12-04
[fix] destroy connection on test teardown

georgehristov:
Fix for Burn test "Too many connections" failure

### DIFF
--- a/src/Oracle/Connection.php
+++ b/src/Oracle/Connection.php
@@ -22,66 +22,6 @@ class Connection extends BaseConnection
         ])->execute();
     }
 
-    // {{{ fix for too many connections for CI testing
-
-    /** @var int */
-    private static $ciDifferentDsnCounter = 0;
-    /** @var array */
-    private static $ciLastConnectDsn;
-    /** @var \PDO */
-    private static $ciLastConnectPdo;
-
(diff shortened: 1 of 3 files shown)


DarkSide666 (approved):
LGTM.
Tests fail because new PHPUnit 9.5 doesn't have `getAnnotations()` method anymore.

mvorisek (changes requested):
in Oracle, we have a special PDO reuse, maybe for this reason, try to remove it

if the tests pass then, then this is probably welcomed

also, what about the another test cases?

georgehristov:
@mvorisek I saw tests failing (on MySQL) due to too many connections and I applied same fix as in data.
Just to describe it in words: the issue is that if a model is stored in test class property on test setUp then the connection is not closed on test tearDown unless this property is unset.
So each test creates one more connection until the connection limit is reached. If number of tests is below that limit then all appears fine.

mvorisek:
I am not arguing it, see my additional commit about that Oracle issue. If these fixed by you solves it now here and in data, than both 👍

mvorisek (approved):

mvorisek:
Oracle is not fixed by this... I think this fix is "correct" in sense of cleaness, ie. what is setup is released too.

PR #282, opened 2020-12-04
[update] add support for Query::orExpr()->having(...)

### LINKED ISSUE #280: Query cannot render complex condition in the HAVING clause
Using `Query::andExpr()` and `Query::orExpr()` only works for WHERE clause

georgehristov:
Query::orExpr()->having(...) does not render the conditions.

This can be overcome by using Query::orExpr()->where(...) but then code is not consistent.

Fixes #280

(Failure caught when creating tests for `\atk4\data\Model\Aggregate`)

### DIFF
--- a/tests/QueryTest.php
+++ b/tests/QueryTest.php
@@ -1537,6 +1537,7 @@ class QueryTest extends AtkPhpunit\TestCase
 
     public function testNestedOrAndHaving()
     {
+        // test 1
         $q = $this->q();
         $q->table('employee')->field(new Expression('sum([])', ['amount']), 'salary')->group('type');
         $q->having(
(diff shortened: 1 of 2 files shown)


mvorisek (approved):
Approving, but 1 test case would be nice.

tests/QueryTest.php
```diff
@@ -1535,6 +1535,22 @@ public function testNestedOrAnd()
...
     }
 
+    public function testNestedOrAndHaving()
+    {
+        $q = $this->q();
+        $q->table('employee')->field(new Expression('sum([])', ['amount']), 'salary')->group('type');
+        $q->having(
+            $q
+                ->orExpr()
+                ->having('a', 1)
```
mvorisek, line 1545: are we safe if `where()` is used here?

georgehristov: Yes, even it has priority

So now  you can use either `where` or `having`. If both of them are used it will take only `where` and disregard `having` to avoid mixing them

mvorisek: I think we should throw instead - when rendering parent where/having, check if the subexpression has not other kind set.

mvorisek: and probably as another test method like `testNestedOrAndHavingWithWhereException`

DarkSide666: Why don't allow to mix `where` and `having`? In SQL it's possible to have:
```
SELECT name, sum(amount) total
FROM t
WHERE deleted=0
GROUP BY name
HAVING total > 100
```

mvorisek: Only when nested! If however that what you have written is not allowed, then add a test and @georgehristov must fix asap :)

georgehristov: @DarkSide666 this is only for the query nested expressions `Query::orExpr()` / `Query::andExpr()` used for building complex WHERE / HAVING. 
These use simple template `[orwhere]` or `[andwhere]` and only implode the conditions in a string with `or` / `and`

In the full query you can of course still use WHERE and HAVING combined.

mvorisek: It would be great if we can rename creator methods like `orExpr` to `createOrExpr`.

PR #283, opened 2020-12-11
Upgrade DBAL to >= 2.10

mvorisek:
we can then use types enum up too the lastest 3.x DBAL version

we can upgrade because ORM 2.8.x was released


georgehristov (approved):
LGTM

PR #287, opened 2020-12-19
Improve PHPStan to level 5

mvorisek:


tests/RandomTest.php
```diff
@@ -21,12 +21,8 @@ public function q(...$args)
         return new Query(...$args);
     }
 
-    /**
-     * @doesNotPerformAssertions
-     */
     public function testMiscInsert()
```
mvorisek, line 24: @georgehristov @DarkSide666 ok to remove this test? I do not see any value in it...

georgehristov (approved):

PR #293, opened 2021-01-19
Fix DBAL Connection to be GCed without an explicit call needed

mvorisek:
fix burn/cron testing with 250 iterations

the issue is in cyclic references that need an explicit GC

### DIFF
--- a/src/Connection.php
+++ b/src/Connection.php
@@ -57,6 +57,14 @@ abstract class Connection
         $this->setDefaults($properties);
     }
 
+    public function __destruct()
+    {
+        // needed for DBAL connection to be released immeditelly
+        if ($this->connection !== null) {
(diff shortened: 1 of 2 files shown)


DarkSide666 (approved):

src/Connection.php
```diff
@@ -57,6 +57,14 @@ public function __construct($properties = [])
         $this->setDefaults($properties);
     }
 
+    public function __destruct()
+    {
+        // needed for DBAL connection to be released immeditelly
+        if ($this->connection !== null) {
+            $this->connection->close();
```
mvorisek, line 64: @georgehristov FYI, DBAL connection is cyclic an explicit GC collect cycles call is needed to release it, I belive this was also the cause for atk4/data issue originally

PR #295, opened 2021-03-15
Fix getDebugQuery with bool param

### LINKED ISSUE #291: Cast bool params to int in getDebugQuery
When executing a query a bool parameter is cast to an int since SQL doesn't really like bool types. In the 'getDebugQuery' function though there is no handling for bool types so you get an output like
```
SELECT 
    *
FROM
    orders
WHERE
    is_paid =  AND user = 1
```

mvorisek:
fixes https://github.com/atk4/dsql/issues/291

### DIFF
--- a/src/Expression.php
+++ b/src/Expression.php
@@ -481,7 +481,7 @@ class Expression implements \ArrayAccess
                 continue;
             }
 
-            $result = preg_replace('~' . $key . '(?=[^_]|$)~', $replacement, $result);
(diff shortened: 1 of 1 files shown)


mvorisek:
@scottfriebel please confirm it fixes your issue

scottfriebel:
This should fix it, thanks!

DarkSide666 (approved):
LGTM

PR #300, opened 2021-05-01
Fix PostgreSQL boolean handling for PHP 8.0.5 II

mvorisek:


mvorisek:
https://github.com/atk4/data/pull/870 fix is enough

